PR #4, opened 2023-10-27
Added deployment workflow and associated config

### LINKED ISSUE #3: Deployment: Send Something from Nirin to Gadi via Workflow
With the runners currently existing on Nirin, we must determine ways to move executables and other information that is created on Nirin to the Gadi system. 

**Considerations**
* How do we move this information? `scp`? Something NCI specific? See: https://forum.access-hive.org.au/t/github-runner-at-nci/1048/3

**Future Considerations**
* What do we move? 
  * Just the executable? What metadata do we need to move? A `spack.lock` file from CI process seems helpful. 
* Where do we move it? 
  * Is it somewhere that is accessible to everyone? Do we need a service user?
  * Do we need to have `spack` installed on Gadi?
* What do we do with it when we've moved it?
  * Something to do with `module load`?
  
References earlier issue ACCESS-NRI/build-ci#5

CodeGat:
Transferred from https://github.com/ACCESS-NRI/build-ci/pull/118

Closes #3

### DIFF
--- a/.github/workflows/deploy-1-setup.yml
+++ b/.github/workflows/deploy-1-setup.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Get Model
         id: get-model
         # for the cases where the repo name is in uppercase but the package name is lowercase (eg. access-nri/MOM5)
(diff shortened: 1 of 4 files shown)


jo-basevi (approved):
LGTM

PR #7, opened 2023-10-31
deploy-2-start.yml: Added ssh setup steps for each job

### LINKED ISSUE #6: Pass SSH Key into GitHub-hosted runner
Reverting back to GitHub Hosted runners for the meantime, which means that we can pass the private key as a secret to a native call to ssh/rsync.

CodeGat:
Ported the ssh code from the access-nri/test project to the deployment workflow. 

Closes #6

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -29,12 +29,23 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.version }}
+      - name: Setup ssh
+        run: |
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          echo '${{ secrets.SSH_KEY }}' > ~/.ssh/priv.key
+          chmod 600 ~/.ssh/priv.key
+          ssh-keyscan ${{ secrets.HOST }} >> ~/.ssh/known_hosts
+          ssh-keyscan ${{ secrets.HOST_DATA }} >> ~/.ssh/known_hosts
+          chmod 600 ~/.ssh/known_hosts 
       - name: Copy spack.yaml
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -29,12 +29,23 @@ jobs:
...
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          echo '${{ secrets.SSH_KEY }}' > ~/.ssh/priv.key
+          chmod 600 ~/.ssh/priv.key
+          ssh-keyscan ${{ secrets.HOST }} >> ~/.ssh/known_hosts
+          ssh-keyscan ${{ secrets.HOST_DATA }} >> ~/.ssh/known_hosts
+          chmod 600 ~/.ssh/known_hosts 
       - name: Copy spack.yaml
-        run: rsync spack.yaml ${{ secrets.USER }}@${{ secrets.HOST_DATA }}:${{ vars.SPACK_YAML_LOCATION }}
+        run: rsync -e 'ssh -i ~/.ssh/priv.key' \
```
jo-basevi, line 42: An alternative to adding to known hosts might to setup an ~/.ssh/config with Hostname, User, IdentityFile etc so subsequent rsync/ssh cmds could just use the Host  rather than `${{ secrets.USER}}@${{ secrets.HOST }} -i ~/.ssh/deploy.priv `. In this case I think it will still add the same about of code so might not be worth it

jo-basevi (approved):
LGTM

.github/workflows/deploy-2-start.yml
CodeGat: Thats a good idea - I'll create an issue for it. Thanks for all your help with the development of this pull request :)

PR #10, opened 2023-11-01
Checkout caller workflow fix

CodeGat:
Removing the args to actions/checkout as fetch-tags isn't needed due to the checkout ref being the value of on.push.tags, and fetch-depth being by default 1


jo-basevi (approved):
LGTM

PR #11, opened 2023-11-02
GitHub Environment name fix

CodeGat:
The name of the environment in build-cd is 'Gadi' not 'gadi', it's instead creating an environment in ACCESS-OM2 named 'gadi'. This PR uses the same capitalization.


jo-basevi (approved):
LGTM

PR #12, opened 2023-11-02
Identity File name standardisation

CodeGat:
The name of the key file used to be 'deploy.priv' but should have been 'priv.key'


jo-basevi (approved):
LGTM

PR #14, opened 2023-11-02
Merge Deployment and Release Jobs

### LINKED ISSUE #13: Merge Deployment and Release Jobs
Multiple Environment-associated Jobs mean that there are technically multiple deployments. This means that you can  approve the deployment, but not approve of the release. This doesn't make sense, so we should merge the two.

CodeGat:
Merged the deployment and release jobs, because I had to approve of the deployment and the release separately. If deployment succeeds, then the creation of a github release should be run automatically. Closes #13.


jo-basevi (approved):
LGTM

PR #15, opened 2023-11-13
Added debug and fresh install flags to `spack install`

CodeGat:
In this PR:
* added --debug flag so we get more output in CD pipeline
* added --fresh flag so install doesn't having clashing `package@git.version=spack_version` tags 

Should close https://github.com/ACCESS-NRI/ACCESS-OM2/issues/15


jo-basevi (approved):
LGTM

PR #23, opened 2023-11-29
CD: Deploy new version of spack to a deployment target

CodeGat:
This PR:
* Creates a new install of spack under `.../vk83/apps/spack/<version>` for a given version
* Clones `spack-packages` and `spack-config` at an optional version
* Links the appropriate configs together

This is a `workflow_dispatch` workflow, which means that it is triggered from GitHub itself. 
Since it is run in a GitHub Environment, it is considered a deployment, and has associated deployment rules. This means that someone from the release team would need to sign off on the deployment.

### DIFF
--- a/.github/workflows/create-deployment-spack.yml
+++ b/.github/workflows/create-deployment-spack.yml
@@ -1,0 +1,52 @@
+name: Create Deployment Spack
+on:
+  workflow_dispatch:
+    inputs:
+      spack-version:
+        type: string
+        required: true
+        description: A version of spack
+      spack-packages-version:
+        type: string
+        required: true
+        default: main
+        description: A version of ACCESS-NRI/spack-packages
+      spack-config-version:
+        type: string
+        required: true
+        default: main
+        description: A version of ACCESS-NRI/spack-config
+jobs:
+  create-spack:
+    name: Spack
+    runs-on: ubuntu-latest
+    # all `secrets` and `vars` inherit from the 'Gadi Spack' GitHub Environment,
+    # which can be found in the Settings of the repository.
+    environment: "Gadi Spack"
+    steps:
+      - name: Setup SSH
+        id: ssh
+        uses: access-nri/actions/.github/actions/setup-ssh@main
+        with:
+          hosts: ${{ secrets.SSH_HOST }}
+          private-key: ${{ secrets.SSH_KEY }}
+
+      - name: Strip Spack Version
+        id: strip
+        # this step removes the 'release/' part from some spack tags, so our
+        # directory structure doesn't contain a 'releases' subdirectory
+        run: echo "version-dir=$(echo '${{ inputs.spack-version }}' | cut --delimiter '/' --fields 2)" >> $GITHUB_OUTPUT
+
+      - name: Install
+        env:
+          ROOT_VERSION_LOCATION: ${{ vars.ROOT_SPACK_LOCATION }}/${{ steps.strip.outputs.version-dir }}
+        # This step will fail if `mkdir` fails to create a version directory (e.g. `0.20` is already deployed and exists)
+        run: |
+          ssh ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          mkdir ${{ env.ROOT_VERSION_LOCATION }} || exit $?
+          git -C ${{ env.ROOT_VERSION_LOCATION }} clone -c feature.manyFiles=true https://github.com/spack/spack.git --branch ${{ inputs.spack-version }} --single-branch --depth=1
+          git -C ${{ env.ROOT_VERSION_LOCATION }} clone https://github.com/ACCESS-NRI/spack-packages.git --branch ${{ inputs.spack-packages-version }}
+          git -C ${{ env.ROOT_VERSION_LOCATION }} clone https://github.com/ACCESS-NRI/spack-config.git --branch ${{ inputs.spack-config-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-deployment-spack.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install
+        env:
+          SPACK_VERSION_LOCATION: ${{ vars.ROOT_SPACK_LOCATION }}/${{ inputs.spack-version }}
+          SPACK_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack
+          SPACK_PACKAGES_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack-packages
+          SPACK_CONFIG_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack-config
+          SPACK_RELEASE_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/release
+        run: |
+          ssh ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          mkdir ${{ env.SPACK_VERSION_LOCATION }}
```
aidanheerdegen, line 40: I believe this will error if the directory already exists, and therefore the job will fail and not attempt subsequent steps. Can you confirm that is the case? I would like it to not work if the directory already exists.

CodeGat: Good point. I believe it will attempt later steps, but will confirm. 
So just to confirm, if the directory exists (AKA, we have already deployed this version of spack), the job should fail.

.github/workflows/create-deployment-spack.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Setup SSH
+        id: ssh
+        uses: access-nri/actions/.github/actions/setup-ssh@main
+        with:
+          hosts: ${{ secrets.SSH_HOST }}
+          private-key: ${{ secrets.SSH_KEY }}
+      - name: Install
+        env:
+          SPACK_VERSION_LOCATION: ${{ vars.ROOT_SPACK_LOCATION }}/${{ inputs.spack-version }}
```
aidanheerdegen, line 33: Took me a while, but I found `vars.ROOT_SPACK_LOCATION` was set in the "Gadi Spack" environment. Would be good to add a comment with this information to make it obvious where to find it, and maybe change it if required.

CodeGat:
I also realised that we would be creating a strange directory structure if the version of spack that we want to clone is something like `releases/0.20`, which would lead to `.../vk83/apps/spack/release/0.20/...` rather than just `...vk83/apps/spack/0.20/...`. 
Will draft and refactor this potential annoyance.

.github/workflows/create-deployment-spack.yml
```diff
@@ -0,0 +1,46 @@
...
+          SPACK_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack
+          SPACK_PACKAGES_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack-packages
+          SPACK_CONFIG_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/spack-config
+          SPACK_RELEASE_LOCATION: ${{ env.SPACK_VERSION_LOCATION }}/release
+        run: |
+          ssh ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          mkdir ${{ env.SPACK_VERSION_LOCATION }}
+          git clone -c feature.manyFiles=true https://github.com/spack/spack.git ${{ env.SPACK_VERSION_LOCATION }} --branch ${{ inputs.spack-version }} --single-branch --depth=1
+          git clone  https://github.com/ACCESS-NRI/spack-packages.git ${{ env.SPACK_VERSION_LOCATION }} --branch ${{ inputs.spack-packages-version }}
+          git clone  https://github.com/ACCESS-NRI/spack-config.git ${{ env.SPACK_VERSION_LOCATION }} --branch ${{ inputs.spack-config-version }}
```
aidanheerdegen: You've specified `--single-branch --depth=1` for `spack` but not the others. I assume this is because of the sheer size of the `spack` repo?

I don't have a problem with them being different, but is there a reason not to use those options with the other repos too?

Speaking of which, how do we update `spack-packages` and `spack-config` if required?

CodeGat: Correct. 

I don't use them because we may want to swap to a different branch (`--single-branch` makes this a bit harder) for spack-packages/config for some reason, whereas with spack we would just create a new `.../spack/<version>` folder. 

For `spack-config`, we have an on-push workflow: https://github.com/ACCESS-NRI/spack-config/blob/main/.github/workflows/update-deployment-repos.yml
For `spack-packages` we don't have that workflow yet...I might have to create it!

.github/workflows/create-deployment-spack.yml
aidanheerdegen: I believe so, because once it is deployed other deployments will use that location and build software (models and dependencies) in there. If we overwrite it it would be ... bad.

Maybe we would like an option to replace a deployment if something bad happens? That would be a (tactical) nuclear option ...

CodeGat: I think if something that bad happened, we would be allowed to use the account to rectify it.

harshula:
Hi @CodeGat , Can we set the directory permissions in this script?

CodeGat:
Sure, we can add that, as well

harshula (approved):
Reviewed in VC.

.github/workflows/create-deployment-spack.yml
aidanheerdegen: Great, thanks.

PR #27, opened 2023-12-04
Generic Github Release Provenance

### LINKED ISSUE #24: In the GitHub release, note the version of `spack-packages` and `spack-config`
These are two repos that may change versions in the future, so we need to note what version each release was using.

CodeGat:
In this PR:
* Updated deployment workflow to use specific versions of spack-packages and spack-config, which is specified in the calling repository. 

Should close #24﻿


jo-basevi (approved):
LGTM

PR #37, opened 2024-02-06
Undeploy Job jq fix

CodeGat:
In this PR:
- Modified `jq` query - we now escape fields that are delimited by `-` using `""`


aidanheerdegen (approved):
Tricksy!

PR #41, opened 2024-02-08
deploy-*.yml: Added ref and updated version inputs

### LINKED ISSUE #40: Prerelease deployment failing because `version` is the deployment build number
Run: https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/7811118670
Failure: https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/7811118670/job/21305686781#step:2:53

We are using the build deployment number as something to checkout rather than the branch.

CodeGat:
In this PR:
* We were using version as a git ref, when that is no longer applicable. This is now split into the `ref` input, and the `version` input. 

References #40

### DIFF
--- a/.github/workflows/deploy-1-setup.yml
+++ b/.github/workflows/deploy-1-setup.yml
@@ -7,10 +7,15 @@ on:
         required: false
         default: release
         description: The type of deployment - either 'release' or 'prerelease'
+      ref:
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM

PR #43, opened 2024-03-23
Generify Undeploy script

### LINKED ISSUE #42: Generify `undeploy-*` workflow
Specifically https://github.com/ACCESS-NRI/build-cd/blob/36f14cab6687432b9a36f5422d33cb1f605ba6e7/.github/workflows/undeploy-1-setup.yml#L10-L24 - we do a bunch of extra work to get the `spack env` name, but this is all assuming that we would only want to undeploy things of the form `<MODEL>-<VERSION_SPECIFIER>` (eg. `access-om2-2024_02_1-*`).

We should make it more generic and make the `inputs.version-pattern` specify everything.

CodeGat:
In this PR:
* undeploy-1-setup.yml: removed job that prepends model name

Closes #42

### DIFF
--- a/.github/workflows/undeploy-1-setup.yml
+++ b/.github/workflows/undeploy-1-setup.yml
@@ -5,24 +5,8 @@ on:
       version-pattern:
         type: string
         required: true
-        description: A wildcard-supported string for version(s) of the model being removed
+        description: A wildcard-supported string of spack environments to remove
 jobs:
-  get-spack-env:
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM. Presumably there is a corresponding change required in the calling workflow?

CodeGat:
Indeed there is - those workflows are linked in the issue

PR #46, opened 2024-04-08
Add Custom Action `validate-repo-version`

CodeGat:
In this PR:
* Added a new action `validate-repo-version` that essentially is functionality pulled out from ACCESS-OM2s deployment logic: https://github.com/ACCESS-NRI/ACCESS-OM2/blob/bf1f97ca75b942dd08506b88197cf0feaa1c694d/.github/workflows/ci.yml#L28-L74
* Added updates to `README.md`

References ACCESS-NRI/ACCESS-OM3#2

### DIFF
--- a/.github/actions/validate-repo-version/action.yml
+++ b/.github/actions/validate-repo-version/action.yml
@@ -1,0 +1,62 @@
+name: Validate Repo Versions
+description: Action that validates that the versions in `config/versions.json` match existing versions in the repository
+inputs:
+  repos-to-check:
+    type: string
+    required: true
+    description: Space-separated list of ACCESS-NRI repositories to check
+outputs:
+  spack-packages-version:
+    value: ${{ steps.versions.outputs.packages }}
+    description: Version of spack-packages from the `config/versions.json`
+  spack-config-version:
+    value: ${{ steps.versions.outputs.config }}
+    description: Version of spack-config from the `config/versions.json`
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v4
+
+    # The next steps checkout the repos to confirm that the versions in
+    # versions.json exist in the repositories.
+    - name: Setup
+      id: versions
+      run: |
+        if [[ "${{ contains(inputs.repos-to-check, 'spack-packages') }}" == "true" ]]; then
+          echo "packages=$(jq --compact-output --raw-output '."spack-packages"' ./config/versions.json)" >> $GITHUB_OUTPUT
+        fi
+
+        if [[ "${{ contains(inputs.repos-to-check, 'spack-config') }}" == "true" ]]; then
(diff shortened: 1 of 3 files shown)


.github/actions/validate-repo-version/action.yml
```diff
@@ -0,0 +1,64 @@
+name: Validate Repo Versions
+description: Action that validates that the versions in `config/versions.json` match existing versions in the repository
+inputs:
+  repos-to-check:
+    type: string
+    required: true
+    description: Space-separated list of ACCESS-NRI repositories to check
```
jo-basevi, line 7: There's really two supported repos for these checks? If someone mis-spells something or adds a gibberish repo to check, I guess nothing wrong would happen (except for the outputs not being set?) Not sure whether this should be a note in the description or a step with a quick check

jo-basevi (dismissed):
Changes look good to me.

As the code checks looks similar for spack-packages/spack-config, would it be less code for the action to be more general and just to check one repo and then output that version? If the schema for `configs/versions.json` fixed across different models (e.g. spack-packages, spack-config) or if there's going to be a low number of repos, it's probably fine.

.github/actions/validate-repo-version/action.yml
CodeGat: This has been reworked to fail when the repo it is given is not in the `versions.json` :+1:

.github/actions/validate-repo-version/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    description: ACCESS-NRI repository to validate associated version in `config/versions.json`
+outputs:
+  version:
+    value: ${{ steps.jq.outputs.version }}
+    description: Version of spack-packages from the `config/versions.json`
+runs:
+  using: composite
+  steps:
+    # Checkout the callers `config/versions.json`
+    - uses: actions/checkout@v4
```
aidanheerdegen, line 16: Is there a situation where we'd want to check this for a version of the model that is not the `HEAD` of `main`?

aidanheerdegen (approved):
LGTM. Just one small tiny question.

.github/actions/validate-repo-version/action.yml
CodeGat: Indeed - and that case is covered. `actions/checkout` will by default have a `ref` that is dependent on the event type. So for `on.pull_request` it will be the PR branch, and on others it will be `main`. Unless you mean user-customisable, in which case - no. But I don't know if it should be.

aidanheerdegen: I'm thinking about the scheduled check case for a configuration, but perhaps that doesn't intersect with this?

In which case merge away!

CodeGat: No, we don't validate that because in order for it to have been merged and tagged at that point, it must have passed that check during PR

PR #47, opened 2024-04-15
CI/CD Consolidation

CodeGat:
An investigation into moving all complexity out of the model repos themselves, and putting it in `build-cd`.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,59 @@
+name: CD
+run-name: ${{ inputs.model }} CD
+# NOTE: This workflow requires:
+# permissions.contents:write
+# secrets:inherit with an appropriate GitHub Environment for deployment in the caller
+on:
+  workflow_call:
+    inputs:
+      model:
+        type: string
+        required: true
+        description: The model that is being tested and deployed
+  # Callers usually have the trigger:
+  # push:
+  #   branches:
+  #     - main
+  #     - backport/*.*
+  #   paths:
+  #     - config/**
+  #     - spack.yaml
+
+env:
+  SPACK_YAML_MODEL_YQ: .spack.specs[0]
+jobs:
+  push-tag:
+    name: Tag Deployment
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    outputs:
+      name: ${{ steps.tag.outputs.name }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Generate Tag
+        id: tag
+        # Get the tag name from the `spack.yaml` that was merged into main, which
+        # is of the form `<model>@git.<version>`.
+        run: echo "name=$(yq '${{ env.SPACK_YAML_MODEL_YQ }} | split("@git.") | .[1]' spack.yaml)" >> $GITHUB_OUTPUT
+
+      - name: Push Tag
+        # NOTE: Regarding the config user.name/user.email, see https://github.com/actions/checkout/pull/1184
+        run: |
+          git config user.name ${{ vars.GH_ACTIONS_BOT_GIT_USER_NAME }}
+          git config user.email ${{ vars.GH_ACTIONS_BOT_GIT_USER_EMAIL }}
+          git tag ${{ steps.tag.outputs.name }} --force
+          git push --tags --force
+
+  deploy-release:
+    name: Deploy Release
+    needs:
+      - push-tag
+    uses: access-nri/build-cd/.github/workflows/deploy-1-setup.yml@main
+    with:
+      ref: ${{ github.ref_name }}
+      version: ${{ needs.push-tag.outputs.name }}
+    secrets: inherit
+    permissions:
+      contents: write
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -1,5 +1,7 @@
 name: Comment Command
 run-name: ${{ inputs.model }} Comment Command
+# NOTE: This workflow requires:
+# permissions.contents:write
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -1,0 +1,114 @@
+name: Comment Command
+run-name: ${{ inputs.model }} Comment Command
+on:
+  workflow_call:
+    inputs:
+      model:
+        type: string
+        required: true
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,199 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      spack-packages-version: ${{ steps.spack-packages.outputs.version }}
+      spack-config-version:  ${{ steps.spack-config.outputs.version }}
+    steps:
+      - name: Validate spack-packages version
+        id: spack-packages
+        uses: access-nri/build-cd/.github/actions/validate-repo-version@main
+        with:
+          repo-to-check: spack-packages
```
aidanheerdegen, line 52: Just want to check this will work if the workflow is called from outside the ACCESS-NRI organisation.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,199 @@
...
+
+      - name: Validate spack-config version
+        id: spack-config
+        uses: access-nri/build-cd/.github/actions/validate-repo-version@main
+        with:
+          repo-to-check: spack-config
```
aidanheerdegen, line 58: As above.

aidanheerdegen (commented):
Looks great!

.github/workflows/ci-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+          versioning-scheme: calver-minor
+          bump-type: ${{ steps.setup.outputs.bump }}
+
+      - name: Update, Commit and Push the Bump
+        run: |
+          git config user.name ${{ vars.GH_ACTIONS_BOT_GIT_USER_NAME }}
+          git config user.email ${{ vars.GH_ACTIONS_BOT_GIT_USER_EMAIL }}
+
+          yq -i '${{ env.SPACK_YAML_MODEL_YQ }} = "${{ inputs.root-sbd }}@git.${{ steps.bump.outputs.after }}"' spack.yaml
+          yq -i '${{ env.SPACK_YAML_MODEL_PROJECTION_YQ }} = "{name}/${{ steps.bump.outputs.after }}"' spack.yaml
```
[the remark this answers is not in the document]
CodeGat: `{name}`, in this context, is a placeholder that `spack` uses for modulefile projection generation. It defaults to the name of the package. For example:
```yaml
# ...
projections:
  all: '{name}/{version}'
  access-om3: '{name}/2024.04.0'
```
`access-om3`s projection will be `access-om3/2024.04.0`

.github/workflows/ci-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+  workflow_call:
+    inputs:
+      model:
+        type: string
+        required: true
+        description: The model that is being tested and deployed
+      root-sbd:
+        type: string
+        required: false
+        default: ${{ inputs.model }}
```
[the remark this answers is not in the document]
CodeGat: Yeah! Just realized that you can access that context in the `inputs` (as well as `vars`, `github`). Very useful.

.github/workflows/ci.yml
CodeGat: Currently, no it won't. This is because we are comparing the version specified in the `versions.json` like so: https://github.com/ACCESS-NRI/ACCESS-OM2/blob/main/config/versions.json#L3
Which doesn't account for a repo owner (we assume it to be ACCESS-NRI).

Would we want it to be used more widely? We would have to update each models `versions.json` to be Something like:
```json
"ACCESS-NRI/spack-packages": "2023.02.11"
```

.github/workflows/ci.yml
CodeGat: Likewise :D

.github/workflows/cd.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: tag
+        # Get the tag name from the `spack.yaml` that was merged into main, which
+        # is of the form `<model>@git.<version>`.
+        run: echo "name=$(yq '${{ env.SPACK_YAML_MODEL_YQ }} | split("@git.") | .[1]' spack.yaml)" >> $GITHUB_OUTPUT
+
+      - name: Push Tag
+        # NOTE: Regarding the config user.name/user.email, see https://github.com/actions/checkout/pull/1184
+        run: |
+          git config user.name ${{ vars.GH_ACTIONS_BOT_GIT_USER_NAME }}
+          git config user.email ${{ vars.GH_ACTIONS_BOT_GIT_USER_EMAIL }}
```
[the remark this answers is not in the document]
CodeGat: I'm actually not too sure. Haven't checked whether inner `actions/checkouts` are cleaned up automatically at the end of an action. But yes, this could be an action as well!

CodeGat:
Just realised that this could be mergable since the existing infrastructure is not modified!

.github/workflows/ci.yml
aidanheerdegen: I was wondering how this might affect the CABLE-LSM/CABLE model if the context it was calling it from couldn't find `spack-packages`. I guess though that the `build-cd` action is being called from the ACCESS-NRI org so it's not an issue?

aidanheerdegen:
Another approach to trying to make the `bash` sections more testable and reusable is to create `bash` functions, pop them in a file (say `functions/thinkofagoodname.sh`) and then make sure those functions are in a file that `bash` will source that file every time it is invoked.

One way to do that is an approach like this:

https://github.com/marketplace/actions/action-setup-bash

which is an action that dumps a function into `.bash_profile` and sets the default `bash` to use `--login` and so source that file.

An example function:

```bash
$ cat extract_version.sh 
extract_version() {

   if [ "$#" -ne 2 ]; then
      echo "Illegal number of parameters"
      return 1
   fi

   local node=$1
   local spackyaml=$2

   echo $(yq "${node} | split(\"@git.\") | .[1]" ${spackyaml})

}
```

Now you could have a bunch of files, one for each function and dump them all into `.bash_profile`, or put them all in a single file.

Could then have some specialist CI that only gets called when files in that subdirectory changes to test the functions.

.github/workflows/ci.yml
CodeGat: It should be able to find it - it's a public repository in our org. If that's what you meant?

aidanheerdegen (approved):
Given that I reviewed it before, and nothing has changed, I'll trust former me.

PR #54, opened 2024-04-19
`Prerelease`: Unique Modulefiles 

CodeGat:
An oversight exists where prereleased models would have the same modulefile name as the soon-to-be-released one. For example, `access-om3/2024.02.1`. This would not change between commits in the PR, which would mean a bunch of deployed prereleases would share the same module name above, with only the first being pointed to. 

This PR adds a step into the deployment workflow in which we modify the version of the root SBD that the model is based on, giving it a unique `<model>/pr<number>-<commit>` style, similar to the `spack env` name. For example, `access-om3/pr12-2`.

We've had to propagate the `root-sbd` input from the `ci` step all the way through the deployment workflow, as the `yq` that overwrites the version may have a different name than the actual name of the model. For example, we used to have `access-om3-virtual` being the root SBD of the `access-om3` model, before it was changed.

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -27,7 +27,13 @@ on:
         type: string
         required: true
         description: The GitHub deployment environment name
-
+      root-sbd:
+        type: string
+        required: false
+        default: ${{ inputs.model }}
+        description: The root SBD that is being used as the modulefile name
+env:
+  SPACK_YAML_MODULEFILE_PROJECTION_YQ: .spack.modules.default.tcl.projections.${{ inputs.root-sbd }}
 jobs:
   deploy-to-environment:
     name: Deploy to ${{ inputs.deployment-environment }}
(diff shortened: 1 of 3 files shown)


aidanheerdegen:
If we added the modules to a different module path (in pre-release) this wouldn't be necessary. Just sayin'

CodeGat:
How so? What would that look like?
Would that solve the issue of deploying multiple `spack.yaml`s with the same modulefile projections, like `{name}/2024.03.0`?

aidanheerdegen:
> How so? What would that look like?

`mkdir /g/data/vk83/prerelease/modules`

I feel like this is all dancing around similar areas as https://github.com/ACCESS-NRI/build-cd/issues/59

Pre-release currently uses `spack`, `spack-config` and `spack-packages` from the release directory. If there was an entire  replicated directory structure under `prerelease` then there would be a `modules` directory and all associated config etc.

> Would that solve the issue of deploying multiple `spack.yaml`s with the same modulefile projections, like `{name}/2024.03.0`?

Uh, no it wouldn't. We could either have a modulefile projection that just points to the most recent prerelease, or we'd have to make these changes anyway. As we're moving to using `module load version` within `payu` as the preferred way of changing model versions it makes sense to have all pre-release deployments available as modules (see https://github.com/ACCESS-NRI/access-om2-configs/issues/77).

So (kinda) ignore my comment above.

.github/workflows/deploy-2-start.yml
```diff
@@ -54,6 +60,14 @@ jobs:
             ${{ secrets.HOST }}
             ${{ secrets.HOST_DATA }}
 
+      - name: Prerelease Modulefile Name
+        if: inputs.type == 'prerelease'
+        # Modifies the name of the prerelease modulefile to the
+        # `pr<number>-<commit>` style. For example, `access-om3/pr12-2`
```
aidanheerdegen, line 66: In the PR text I thought the name was of the form `access-om3-pr12-2`

CodeGat: In the PR description it has:
> For example, `access-om3/pr12-2`.

aidanheerdegen (approved):
LGTM

PR #60, opened 2024-04-26
Validate `spack.yaml` against our own restricted subset of `spack/spack`s schema

### LINKED ISSUE #55: Investigate whether we should have a `yaml` schema for our flavour of `spack.yaml`
Our `spack.yaml`s seem to have a particular shape. They all so far require:
* a single line in the `.spack.specs` section, where the root SBD resides
* customisation of the packages and dependencies from within the `.spack.packages` section
* some of the versions that are a projection must match the versions that we specify in the `.spack.[packages|specs]` section 

Maybe we should validate this with a proper tool, similar to what we do with the `jsonschema` stuff. Then we can point people to their errors are, if we have a particular format. It'd simplify all the checks we do on the `spack.yaml` in the CI, too. 

Pinging @aidanheerdegen as well

CodeGat:
In this PR:
* Added `validate-spack-yaml` action
* Referenced it in `ci.yml`

Closes #55

### DIFF
--- a/.github/actions/validate-spack-yaml/action.yml
+++ b/.github/actions/validate-spack-yaml/action.yml
@@ -1,0 +1,43 @@
+name: Validate ACCESS-NRI-restrictions on spack.yaml
+description: Action that validates ACCESS-NRI's own spack.yaml format to allow for seamless CI/CD between models
+inputs:
+  schema-version:
+    type: string
+    required: true
+    description: Version of the schema required in SchemaVer
+  schema-repo:
+    type: string
+    required: false
+    default: access-nri/schema
+    description: OWNER/REPO format for the repository containing the schema
+  schema-location:
+    type: string
+    required: false
+    default: au.org.access-nri/model/spack/environment/deployment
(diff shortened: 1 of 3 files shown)


.github/actions/validate-spack-yaml/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      # Checkout the callers `spack.yaml`
+    - uses: actions/checkout@v4
+
+    # Checkout the `schema-repo`
+    - uses: actions/checkout@v4
+      with:
+        repository: ${{ inputs.schema-repo }}
+        path: schema
+
+    - uses: GrantBirki/json-yaml-validate@v2.7.1
```
aidanheerdegen, line 35:
```suggestion
    - uses: GrantBirki/json-yaml-validate@d7814b94473939c1daaca2c96131b891d4703a3c # v2.7.1
```

Hate to be *that* guy ....

aidanheerdegen (changes requested):
So for the pedantry

aidanheerdegen (approved):
LGTM

PR #61, opened 2024-04-30
`ACCESS-OM3` `esmf` OOM Parallelism Fix

CodeGat:
The deployment of ACCESS-OM3 is failing because it is being killed by `Gadi`s OOMKiller when installing `esmf`. This PR explicitly turns down the parallelism to `--jobs 4` to reduce the memory used. 

See the failures here: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/5

Closes ACCESS-NRI/ACCESS-OM3#8

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -78,17 +78,25 @@ jobs:
         # ssh into deployment environment, create and activate the env, install the spack.yaml.
         run: |
           ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          # Update spack-packages/config
           git -C ${{ vars.SPACK_PACKAGES_LOCATION }} fetch
           git -C ${{ vars.SPACK_PACKAGES_LOCATION }} checkout --force ${{ steps.versions.outputs.packages }}
           git -C ${{ vars.SPACK_CONFIG_LOCATION }} fetch
           git -C ${{ vars.SPACK_CONFIG_LOCATION }} checkout --force ${{ steps.versions.outputs.config }}
+
+          # Enable spack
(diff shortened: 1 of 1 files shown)


CodeGat:
Alternatively, we could have a `vars.SPACK_INSTALL_PARALLEL_JOBS` var that can be set to `--jobs 4` (or not set at all) leading to at least a difference in parallel jobs on a per-model basis. I think that might be better (although it would lead to another thing in the `vars` context

CodeGat:
For those with access, I've set `vars.SPACK_INSTALL_PARALLEL_JOBS` in ACCESS-OM3 https://github.com/ACCESS-NRI/ACCESS-OM3/settings/environments/2707603294/edit to `--jobs 4`. It is unset in ACCESS-OM2, and shouldn't affect the `spack install` command, since an unset `vars` expands to ''

.github/workflows/deploy-2-start.yml
```diff
@@ -78,17 +78,25 @@ jobs:
...
           git -C ${{ vars.SPACK_CONFIG_LOCATION }} checkout --force ${{ steps.versions.outputs.config }}
+
+          # Enable spack
           . ${{ vars.SPACK_CONFIG_LOCATION }}/spack-enable.bash
+
+          # Create environment and build model
           spack env create ${{ inputs.env-name }} ${{ vars.SPACK_YAML_LOCATION }}/spack.yaml
           spack env activate ${{ inputs.env-name }}
-          spack --debug install --fresh || exit $?
+          spack --debug install --fresh ${{ vars.SPACK_INSTALL_PARALLEL_JOBS }} || exit $?
```
aidanheerdegen, line 93: Nice work-around

aidanheerdegen (approved):
Appreciate the additional inline comments. They do assist in scaffolding understanding and comprehension.

PR #63, opened 2024-05-02
Upload Build Provenance Metadata

### LINKED ISSUE #33: Report released builds to DB for provenance
For complete experiment provenance there is a requirement to connect executables used in a model with the code and `spack` build for that executable.

If the executable could report its build provenance then `payu` could capture that information directly. I don't think this is possible though

Currently the experiment metadb tracks the executables used in a `payu` model run through the manifest file, which stores a `md5` hash of the executable file. 

If the build-cd workflow uploaded information about a released build to a database, and included the `md5` hash of the executables, we could make the necessary connection between a model run, the executables used and build provenance of those executables.

CodeGat:
In this PR:
* Addition of hashes to `spack.location.json`
* Addition of new repo `secrets`:  `BUILD_DB_CONNECTION_STR`
* Addition of new `vars`: `BUILD_DB_PACKAGES`
* Move non-deploy logic out of `deploy-to-environment` job
* Added tests for new scripts

Closes #33


scripts/upload-build-metadata.bash
```diff
@@ -0,0 +1,71 @@
...
+    --argjson component "$component" \
+    '{
+      component_build: $component,
+      model_build: $model
+    }' > package.json
+
+  cat package.json
+
+  python ./tools/release_provenance/save_release.py package.json
+  echo "Uploaded $pkg to build DB"
```
aidanheerdegen, line 70: I think it is better to do the upload as a command in the CD workflow (maybe && with invoking this script).

In which case maybe this would be renamed to generate_build_metadata.bash or similar?

By doing it this way the script can be tested independently of upload. Speaking of which is there value in adding a test for this into the repo?

CodeGat: Makes sense to me - I'll rename it and move the packaging logic out once the schema is ready to go

aidanheerdegen: It makes it a lot easier to review if there is some test data to run it against.

CodeGat: And also the test data stuff too, will add that as well

CodeGat: Added in https://github.com/ACCESS-NRI/build-cd/pull/63/commits/4e89c561b90956cc30baa5539ba97aec68865024

tests/scripts/generate-build-metadata/valid/cice5.json
aidanheerdegen, line 1: Don't we want a single `json` file for each model, with a single `model_build` object and multiple `model_build` objects?

aidanheerdegen (changes requested):
Sorry!

tests/scripts/generate-build-metadata/valid/cice5.json
CodeGat: If so, we might have to update the release script? @utkarshgupta95 does the release script support something like this at the moment?

```json
{
  model_build: { ... },
  "component_build": {
    "spec": "mom5@git.2023.10.19=2023.10.19",
    ....
  },
  "component_build": {
    "spec": "cice5@git.2024.04.0",
    ...
  }
}
```

tests/scripts/generate-build-metadata/valid/cice5.json
```diff
@@ -0,0 +1,17 @@
+{
+  "component_build": {
+    "spack_hash": "v3zncpqjj2gyseudbwiudolcjq3k3leo",
+    "spec": "cice5@git.2023.10.19=2023.10.19",
+    "install_path": "/g/data/vk83/apps/spack/0.20/release/linux-rocky8-x86_64/intel-19.0.5.281/cice5-git.2023.10.19=2023.10.19-v3zncpqjj2gyseudbwiudolcjq3k3leo",
+    "created_at": "2024-03-27T05:29:36Z",
+    "release_url": "http://example.org/releases"
```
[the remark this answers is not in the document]
utkarshgupta95: I will make this change now

tests/scripts/generate-build-metadata/valid/cice5.json
utkarshgupta95: I will update the script to have component_build array

aidanheerdegen (approved):
LGTM! Let's give this a burl!

PR #65, opened 2024-05-07
added release provenance script

utkarshgupta95:
Script to upload build release information to be utilised in deployment pipeline.

Part of addressing #33

### DIFF
--- a/tools/release_provenance/models.py
+++ b/tools/release_provenance/models.py
@@ -1,20 +1,18 @@
+import os
 from sqlalchemy import (
     DateTime, Text, String, Column, ForeignKey )
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship, sessionmaker
 
-USERNAME = ""
-PASSWORD = ""
-HOST = "ceqv5w67lzc.db.cloud.edu.au"
-DATABASE = "TEST_RELEASE_PROVENANCE"
-CONNSTR = f'postgresql://{USERNAME}:{PASSWORD}@{HOST}/{DATABASE}'
-
 Base = declarative_base()
 
 def create_session():
+    connection = os.getenv("BUILD_DB_CONNECTION_STR")
+    if connection is None:
+        raise Exception("No BUILD_DB_CONNECTION_STR found, check model repository secrets")
 
-    engine = create_engine(CONNSTR)
+    engine = create_engine(connection)
     Base.metadata.create_all(engine)
 
     Session = sessionmaker(bind=engine, autoflush=False)
--- a/tools/release_provenance/models.py
+++ b/tools/release_provenance/models.py
@@ -1,0 +1,50 @@
+from sqlalchemy import (
+    DateTime, Text, String, Column, ForeignKey )
+from sqlalchemy import create_engine
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship, sessionmaker
+
+USERNAME = ""
+PASSWORD = ""
+HOST = "ceqv5w67lzc.db.cloud.edu.au"
+DATABASE = "TEST_RELEASE_PROVENANCE"
+CONNSTR = f'postgresql://{USERNAME}:{PASSWORD}@{HOST}/{DATABASE}'
+
+Base = declarative_base()
+
+def create_session():
+
+    engine = create_engine(CONNSTR)
+    Base.metadata.create_all(engine)
+
+    Session = sessionmaker(bind=engine, autoflush=False)
+    return Session()
+
+
+class ComponentBuild(Base):
+    __tablename__ = "component_build"
+
+    spack_hash = Column(String, primary_key=True, index=True)
+    spec = Column(String, nullable=False)
+    install_path = Column(String, nullable=False, unique=True)
+    created_at = Column(DateTime, nullable=False)
+    release_url = Column(Text, nullable=False, unique=True)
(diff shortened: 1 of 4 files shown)


tools/release_provenance/models.py
```diff
@@ -0,0 +1,50 @@
...
+    DateTime, Text, String, Column, ForeignKey )
+from sqlalchemy import create_engine
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship, sessionmaker
+
+USERNAME = ""
+PASSWORD = ""
+HOST = "ceqv5w67lzc.db.cloud.edu.au"
+DATABASE = "TEST_RELEASE_PROVENANCE"
+CONNSTR = f'postgresql://{USERNAME}:{PASSWORD}@{HOST}/{DATABASE}'
```
CodeGat, line 11: Can these (or just the `CONNSTR`) be passed into the script via `stdin` instead of being constants?

tools/release_provenance/test_release_data.json
```diff
@@ -0,0 +1,16 @@
+{
+    "spack_hash": "ewcdbrfukblyjxpkhd3mfkj4yxfolal4",
+    "spec":  "mom5@git.2023.11.09=2023.11.09",
+    "model_build": {
+        "spack_hash": "f4f2qe5b3c22q4dtzzqs5o2iygy435re",
+        "spec": "access-om2@git.2023.11.23=2023.11.23",
```
CodeGat, line 6: How does this bit work? Is this the spec/spack_hash of the model that contains the components? What happens if we need an entry for `access-om3` or `access-om2` rather than the components?

CodeGat (changes requested):
Just some comments

tools/release_provenance/models.py
CodeGat: Just so we can have something like `save_release.py my.json ${{ secrets.BUILD_DB_CONNECTION_STR }}`

utkarshgupta95: I don't think it's a good idea to pass cred with `stdin`. As they are constants we can directly fetch the value from secrets in the script itself. One such example would be to use `os.environ['const_name']`. I haven't done that bc I wasn't sure which method you would prefer to access these constants or the naming convention you wanna follow in github secrets. I feel it is better if I leave it for you to edit later.

CodeGat: Updated it so the secret is set on the runner and obtained via `BUILD_DB_CONNECTION_STR`

CodeGat (dismissed):

tools/release_provenance/models.py
```diff
@@ -1,20 +1,18 @@
...
-USERNAME = ""
-PASSWORD = ""
-HOST = "ceqv5w67lzc.db.cloud.edu.au"
-DATABASE = "TEST_RELEASE_PROVENANCE"
-CONNSTR = f'postgresql://{USERNAME}:{PASSWORD}@{HOST}/{DATABASE}'
-
 Base = declarative_base()
 
 def create_session():
+    connection = os.getenv("BUILD_DB_CONNECTION_STR")
```
aidanheerdegen, line 11: Could you just add a comment about what form the connection string should take, e.g. just a made-up example. Secrets aren't viewable, so if you needed to change it would be good to have some idea about what it should look like.

aidanheerdegen (approved):
One comment, otherwise looks good to me.

tools/release_provenance/models.py
CodeGat: Sure.

aidanheerdegen (dismissed):
LGTM

CodeGat (changes requested):
Just awaiting @utkarshgupta95 s changes to the schema

CodeGat (approved):

tools/release_provenance/models.py
```diff
@@ -0,0 +1,61 @@
...
+    Session = sessionmaker(bind=engine, autoflush=False)
+    return Session()
+
+
+class ComponentBuild(Base):
+    __tablename__ = "component_build"
+
+    spack_hash = Column(String, primary_key=True, index=True)
+    spec = Column(String, nullable=False)
+    install_path = Column(String, nullable=False, unique=True)
```
CodeGat, line 28: @aidanheerdegen would the install path be nessecarily unique? I suppose it's a similar issue to the spack_hash uniqueness problem.

aidanheerdegen: Given it's on a filesystem and two things named the same thing can't exist in the same location, and assuming we're not overwriting (which we shouldn't be), then yeah I think it should be unique.

tools/release_provenance/models.py
```diff
@@ -26,7 +26,7 @@ class ComponentBuild(Base):
     spec = Column(String, nullable=False)
     install_path = Column(String, nullable=False, unique=True)
     created_at = Column(DateTime, nullable=False)
-    release_url = Column(Text, nullable=False, unique=True)
+    release_url = Column(Text, nullable=False)
```
aidanheerdegen, line 30: This was changed to non-unique just in case? Seemed an unnecessary constraint? Or one we didn't care too much about enforcing?

aidanheerdegen (approved):
LGTM

tools/release_provenance/models.py
CodeGat: Release URLs for components (mom5, cice5,...) will point to the same overall model release (eg. https://github.com/ACCESS-NRI/ACCESS-OM2/releases/tag/2024.03.0) which will be non-unique, so I removed that constraint.

PR #68, opened 2024-05-13
Model names replace `.` with `p` for `ACCESS-ESM1.5`

### LINKED ISSUE #67: `access-esm1.5` model repository name cannot be turned into a `spack env`-compliant name
See:
* https://github.com/ACCESS-NRI/build-cd/blob/c770990f80e95835b4adb2c188b176191e89c107/.github/workflows/ci-closed.yml#L37
* https://github.com/ACCESS-NRI/build-cd/blob/c770990f80e95835b4adb2c188b176191e89c107/.github/workflows/deploy-1-setup.yml#L39

It isn't sufficient to just use the `${{ github.event.repository.name }}` as the spack env name. For example, `spack` doesn't allow `.` in the environment name, which means `access-esm1.5` would fail to create the associated environment.  

Consider further sanitizing the repo name by either:
* replacing the `.` with `_` (consistent with versioning `2024.01.1` -> `2024_01_1`) so `access-esm1.5` -> `access-esm1_5`, or
* replacing the `.` with `p` (consistent with our own spack package naming scheme) so `access-esm1.5` -> `access-esm`1p5`

CodeGat:
This PR was created because `ACCESS-NRI/ACCESS-ESM1.5` contains a `.`, which is not `spack env` name compliant. Since we get the model name (and `spack env` name) from the repo name, we need to do further santization. 

In this PR:
* Each model name sourced from the repository name now lowercases and replaces `.` with `p`. 

Closes #67

### DIFF
--- a/.github/workflows/ci-closed.yml
+++ b/.github/workflows/ci-closed.yml
@@ -34,8 +34,8 @@ jobs:
         id: version
         # For example, `access-om3-pr12-*`
         run: |
-          repo_name_lower=$(echo ${{ github.event.repository.name }} | tr [:upper:] [:lower:])
-          echo "pattern=${repo_name_lower}-pr${{ github.event.pull_request.number }}-*" >> $GITHUB_OUTPUT
+          repo_name_sanitized=$(echo ${{ github.event.repository.name }} | tr [:upper:] [:lower:] | tr '.' 'p' )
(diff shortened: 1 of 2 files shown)


harshula (approved):

harshula:
Please document these types of substitutions in the README.

CodeGat:
Thanks @harshula, noted in a new issue because the `README.md` is more out of date than I thought.

PR #75, opened 2024-05-16
`release_provenance`: Changed requirement to `psycopg2-binary`

CodeGat:
When running the old `requirements.txt` on my local computer, I get an error, even though it works on the runner. 
When I convert to the binary version, it works on the local computer, so I'm updating the `requirements.txt`.

### DIFF
--- a/tools/release_provenance/requirements.txt
+++ b/tools/release_provenance/requirements.txt
@@ -1,2 +1,2 @@
 sqlalchemy==1.4.49
-psycopg2==2.9.9
+psycopg2-binary==2.9.9


aidanheerdegen:
FYI for anyone following along at home, or wondering about this change in the future (including me)

From https://pypi.org/project/psycopg2-binary/

> You can also obtain a stand-alone package, not requiring a compiler or external libraries, by installing the [psycopg2-binary](https://pypi.org/project/psycopg2-binary/) package from PyPI:
> ```bash
> $ pip install psycopg2-binary
> ```
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

aidanheerdegen (approved):

CodeGat:
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

Now I'm not sure I should be using the binary. Confused as to why it installs fine on Gadi without the binary, but has an issue locally. See successful installation on Gadi here: https://github.com/codegat-test-org/test/actions/runs/9024644151/job/24798911202#step:5:17

I might close this PR.

PR #81, opened 2024-05-22
`spack.yaml` existence check

CodeGat:
See failed run: https://github.com/ACCESS-NRI/ACCESS-ESM1.5/actions/runs/9183697365/job/25254826993?pr=5#step:4:20 - this was because it attempts to check for a modification to the spack.yaml SBD version, but the spack.yaml hadn't existed on the base branch yet.

In this PR:
* ci.yml: Add check that continues the CI if there is no spack.yaml to check against on the base branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,13 @@ jobs:
         id: version
         run: |
           git checkout ${{ github.base_ref }}
+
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM as long as we check we end up with a "proper" initial version elsewhere.

PR #83, opened 2024-05-23
Non-`@git.` Version Fix

### LINKED ISSUE #82: Comparing `spack.yaml` versions without a `@git.` attrribute fails
See failed run: https://github.com/ACCESS-NRI/ACCESS-ESM1.5/actions/runs/9199869053/job/25305428602?pr=5

In the following error, we are comparing `null` with a given version. This is because the splitting of the version string (in this case `@access-esm1.5`) expects the format to be `@git.`. Originally, we expected versions to always use the `@git` syntax, but `access-esm1.5` broke that trend. 

Update the splitting logic so it takes the version after `@git.` or `@` (if there's no `git`).

CodeGat:
More detail in the linked issue, but essentially we are going from a string-based split in `yq` to a regex-based match. This is because we want to split on two possible conditions - we want the thing after `@git.` or just `@` if there is no following `git.`. 

Explanatory note: `match(...)` in `yq` returns a kind of yucky bit of yaml that we have to further interrogate to get the actual captured string. For example:
```yaml
string: '@access-esm1.5'
offset: 0
length: 14
captures:
  - string: access-esm1.5
    offset: 1
    length: 13
```
Hence the `captures[0].string` bit. 

Further explanatory note: we have to quote the `"` because functions only accept `"..."` arguments, not `'...'`. But we have to use `"..."` to quote the entire `yq` filter string because we want to interpret the `$DEP` variable. And we have to do that because `yq` doesn't have the nice syntax for variable substitution like `jq` does (`jq --arg name $var ...` is much better).  

In this PR:
* ci.ymls `check-spack-yaml` job: Updated splitting logic, added better error logs, added TODO

Closes #82

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,7 @@ jobs:
         # this step checks that the versions of the packages themselves match with the
         #  names of the modules. For example, access-om3@git.2023.12.12 matches with the
         #  modulefile access-om3/2023.12.12 (specifically, the version strings match)
+        # TODO: Move this into the `scripts` directory - it's getting unweildly.
         run: |
           FAILED='false'
           DEPS=$(yq ".spack.modules.default.tcl.include | join(\" \")" spack.yaml)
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
LGTM

Issue #85, opened 2024-05-24
`Prerelease 0.21` cannot install any models 


CodeGat:
See https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/9203230162/job/25355101630#step:7:91, https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/9216320542/job/25356375475

When deploying onto `Prerelease 0.21`, we get errors relating to the initial checking of the spack install itself:

```
'/bin/git' '-c' 'advice.detachedHead=false' 'fetch' '--tags'
ProcessError: Command exited with status 1:
```

@aidanheerdegen had mentioned that this was an issue when we were `--single-branch` cloning our initial `spack` environments, but `git status` shows that we are currently on the `releases/0.21` branch, not in a `Detached HEAD` state.

Verified that this happens for both existing and new PRs of models, so this issue holds up deployments of models more generally. It was working earlier that day (23/05/2024), see https://github.com/ACCESS-NRI/ACCESS-ESM1.5/pull/5#event-12902392543. 

## Things to do
- [x] Verify that the `spack config blame configs` seems alright 
- [x] Verify that `ACCESS-NRI/spack` hasn't vanished
- [x] Verify that `.git` seems okay in `spack` 
- [x] :sob:
- [x] Determine if it is fetching the tags of something else - not necessarily the `spack` install. Maybe it is actually fetching the tags of `access-om2`, perhaps?
- [x] figure out if the new SSH key on `tm70_ci` is to blame

If all else fails:
* Delete prerelease environment, intialise prerelease environment, redeploy all open model PRs

CodeGat:
When doing a `git log` I note only one commits shows up, and it is grafted:
```
[tm70_ci@gadi-login-05 spack]$ git log
commit f17445a582bb076993f88fef0e0e121a2062a816 (grafted, HEAD -> releases/v0.21, origin/releases/v0.21)
Author: Harshula Jayasuriya <harshula.jayasuriya@anu.edu.au>
Date:   Wed Apr 17 16:15:49 2024 +1000

    fms: add large_file and internal_file_nml variants
    
    * Micael will submit this patch upstream
```
When I run `git tags --fetch` on `spack` itself, it runs fine and has a `0` exit code.

CodeGat:
[...]

PR #86, opened 2024-05-24
Remove 'Rolling Tag' logic for `Prereleases`

CodeGat:
We are removing the rolling tag logic because it is causing `spack install`s to fail. This is because when updating the repository in `spack`, it fails to `git fetch --tags` because it would clobber the existing tags. 
Removes functionality introduced in the CI (back when it was a part of the model) in https://github.com/ACCESS-NRI/ACCESS-OM2/pull/47

In this PR:
* ci.yml: Removed shifting tag logic for prereleases, now only do it initially. 

References #85


aidanheerdegen (approved):
LGTM assuming we still tag again on merge.

CodeGat:
Indeed we do. That is handled in `cd.yml`, specifically here: https://github.com/ACCESS-NRI/build-cd/blob/06ed17bd80bc70941f7be4e202f03581069b7655/.github/workflows/cd.yml#L25-L47 :)

PR #87, opened 2024-05-28
Update regex for extracting a version from `spack.packages.*.require`

CodeGat:
See related ACCESS-NRI/spack-packages#111

This PR updates the regex used to get a version out of a `spack.packages.*.requires` section of a `spack.yaml`. 
We need to update this because the linked issue uses a new `@git.TAG=VERSION` syntax that we don't yet support. 

This would mean that we current support three 'styles' of version:
* `@VERSION` (eg. `@0.2.1` -> `0.2.1`)
* `@git.VERSION` (eg. `@git.2023.12.12` -> `2023.12.12`)
* `@git.TAG=VERSION` (eg. `@git.2024.04.04=access-esm1.5` -> `2024.04.04`)


jo-basevi (approved):
LGTM

PR #91, opened 2024-06-02
Move Long inline scripts into `scripts/`

### LINKED ISSUE #84: Pull out long inline scripts from `run:` steps, add testing
similar to https://github.com/ACCESS-NRI/build-cd/blob/main/scripts/generate-build-metadata.bash which has had a large script pulled out and tested, we need to do this for more scripts. This will make it easier to test independently. This should be the case for all long-running scripts. 

Some candidates:
* https://github.com/ACCESS-NRI/build-cd/blob/06ed17bd80bc70941f7be4e202f03581069b7655/.github/workflows/ci.yml#L132-L156
* https://github.com/ACCESS-NRI/build-cd/blob/06ed17bd80bc70941f7be4e202f03581069b7655/.github/workflows/ci-comment.yml#L51-L84
* I'm sure there are others...

CodeGat:
Some inline scripts would benefit from being able to be run independently of the CI. This also makes them independently testable. 

Attempted to move the actual Gadi deploy logic into a script, but it ended up being a 11 parameter call and was actually less readable. 

Todo:
 - [x] Move scripts
 - [ ] Add basic tests

In this PR:
* Pull out version bumping setup into `scripts/ci-comment`
* Pull out spack.yaml checking into `scripts/ci`
* Move build metadata script into `scripts/deploy-2-start`

Closes #84


CodeGat:
This should probably be closed as it would be more work to get it inline with current CI. Will reopen when I have time.

PR #96, opened 2024-06-26
Deploy to a `spack` specified by `.spack` in `config/versions.json`

### LINKED ISSUE #93: Utilize the `.spack` field from a models `config/versions.json` to determine deployment location
Currently we need to manually switch deployment location in the `vars` of the model repository. Instead, make this dependent on the value in the models `config/versions.json`.

CodeGat:
Instead of having to create `vars` for each full `spack`, `spack-packages` and `spack-config` path on the deployment target, have a single `vars.SPACK_INSTALLS_ROOT_LOCATION` and piecewise construct the above paths using the `.spack` version field in `config/versions.json`. This also means that we don't have to update the above vars when we want to deploy to a different version of spack, and users will have the freedom to pick a deployed version of spack to pre/release to. 

For example, for `spack-packages` version, the full path change from a `vars.SPACK_PACKAGES_LOCATION` to `vars.SPACK_INSTALLS_ROOT_LOCATION/<spack version>/spack-packages`.

 - [ ] Will need to update `vars.SPACK_INSTALLS_ROOT_LOCATION` in our existing `Gadi [Prerelease]` in `ACCESS-OM2`, `ACCESS-OM2-BGC`, `ACCESS-OM3`, `ACCESS-ESM1.5` and then remove `vars.SPACK_*_LOCATION` when it's merged. 

In this PR:
* `deploy-2-start.yml`: Deploy paths created in step using .spack version

Closes #93

### DIFF
--- a/.github/workflows/undeploy-2-start.yml
+++ b/.github/workflows/undeploy-2-start.yml
@@ -26,16 +26,11 @@ jobs:
 
       - name: Get ${{ inputs.deployment-environment }} Remote Paths
         id: path
-        run: |
-          if [ -z "${{ vars.SPACK_INSTALLS_ROOT_LOCATION }}" ]; then
-            echo '::error::`vars.SPACK_INSTALLS_ROOT_LOCATION` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment'
-            exit 1
-          fi
-
-          root=${{ vars.SPACK_INSTALLS_ROOT_LOCATION }}/${{ steps.versions.outputs.spack }}
-          echo "root=$root" >> $GITHUB_OUTPUT
-          echo "spack=$root/spack" >> $GITHUB_OUTPUT
-          echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
+        uses: access-nri/build-cd/.github/actions/get-deploy-paths@main
+        with:
+          spack-installs-root-path: ${{ vars.SPACK_INSTALLS_ROOT_LOCATION }}
+          spack-version: ${{ steps.versions.outputs.spack }}
+          deployment-environment: ${{ inputs.deployment-environment }}
 
       - name: Setup SSH
         id: ssh
--- a/.github/workflows/undeploy-2-start.yml
+++ b/.github/workflows/undeploy-2-start.yml
@@ -16,6 +16,27 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ inputs.deployment-environment }}
     steps:
+      - uses: actions/checkout@v4
+
+      - name: Get Versions From config/versions.json
+        id: versions
+        run: |
+          echo "spack=$(jq --compact-output --raw-output '.spack' ./config/versions.json)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


.github/workflows/undeploy-2-start.yml
```diff
@@ -16,6 +16,27 @@ jobs:
...
+          if [ -z "${{ vars.SPACK_INSTALLS_ROOT_LOCATION }}" ]; then
+            echo '::error::`vars.SPACK_INSTALLS_ROOT_LOCATION` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment'
+            exit 1
+          fi
+
+          root=${{ vars.SPACK_INSTALLS_ROOT_LOCATION }}/${{ steps.versions.outputs.spack }}
+          echo "root=$root" >> $GITHUB_OUTPUT
+          echo "spack=$root/spack" >> $GITHUB_OUTPUT
+          echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
+
```
aidanheerdegen, line 39: Sorry to be *that* guy, but isn't this a copy of the code from `deploy-2-start.yml`? Could it be defined once and reused?

aidanheerdegen:
In case I forget again in the future, the `vars.` namespace is populated in the Settings tab for each repo, e.g. in https://github.com/ACCESS-NRI/ACCESS-OM2/settings/variables/actions

Screenshot 2024-07-01 at 3 14 16 pm

aidanheerdegen (approved):
I know it's not strictly speaking part of this PR, but I'm a bit concerned `vars.SPACK_YAML_LOCATION` is set to be the same for all models. So they could overwrite that path. Why not append a model specific subdirectory to `vars.SPACK_YAML_LOCATION`, or also generate that automatically and put it somewhere that is visible to other users?

.github/workflows/undeploy-2-start.yml
CodeGat: I might make it an action - that would be the most reusable thing

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/96/commits/64193d6bff2008c5170e95d0e565de7ea09044d6

CodeGat:
> I know it's not strictly speaking part of this PR, but I'm a bit concerned `vars.SPACK_YAML_LOCATION` is set to be the same for all models. So they could overwrite that path. Why not append a model specific subdirectory to `vars.SPACK_YAML_LOCATION`, or also generate that automatically and put it somewhere that is visible to other users?

@aidanheerdegen that's a fair concern. It's more of a temporary staging area for the `spack.yaml` before we run `spack env create <env> ${{ vars.SPACK_YAML_LOCATION }}`, where it is then copied into `$SPACK_ROOT/var/spack/environments/<env>/spack.yaml`. I'll add a commit here that disambiguates them - maybe so it is like `<model>.spack.yaml` (for example, `access-om2.spack.yaml`).

aidanheerdegen (approved):
Looks great. The addition of the action is good, not just from a duplication point of view, but I found the documentation really helpful to understand what it was doing.

PR #97, opened 2024-06-26
Can't create a release due to erroneous `./` in artifact upload location

CodeGat:
See failed run: https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/9673293440/job/26688521873#step:4:19
This is due to creating a release that searches for files in `./${{ env.METADATA_PATH }}/*`. However, `env.METADATA_PATH` is set to `/opt/metadata` which means that the final path is invalid: `.//opt/metadata/*`. 

In this PR:
* deploy-2-start.yml: Removed prepended current directory for Release files


jo-basevi (approved):
LGTM

PR #101, opened 2024-06-28
Removed `@git.VERSION` from Prerelease deployment

### LINKED ISSUE #100: `Prerelease`s shouldn't use an `@git.VERSION`
See failed deployment: https://github.com/ACCESS-NRI/ACCESS-OM2-BGC/actions/runs/9705045478/job/26786406673

This failure to `git fetch` is due to a moved tag in `ACCESS-OM2-BGC` (when we move a prerelease to a release):
```
From https://github.com/ACCESS-NRI/ACCESS-OM2-BGC
 * branch            HEAD       -> FETCH_HEAD
 ! [rejected]        2024.03.1  -> 2024.03.1  (would clobber existing tag)
```

We do this initial tagging of a prerelease because currently we deploy the `spack.yaml` as it is - with the `MODEL@git.VERSION` intact. Then, when we move from Prerelease to Release we shift the tag, which leads to the above. 

Instead, we should remove the `@git.VERSION` from the `.spack.specs` for Prereleases.

CodeGat:
Removes the `@git.VERSION` of a `MODEL@git.VERSION` in the `.spack.specs[0]` section of a `spack.yaml` during a Prerelease. This is so we don't have to move tags between Prerelease and Release. 

Tested the `yq` command locally using a `spack.yaml`. 

In this PR:
* `deploy-2-start.yml`: Now removes the `@git.VERSION` section from Prerelease deploy
* `ci.yml`: No longer needs to do the initial tagging of a prerelease

Closes #100

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
     with:
       type: prerelease
       ref: ${{ github.head_ref }}
-      version: ${{ needs.version-tag.outputs.prerelease }}
+      version: ${{ needs.version.outputs.prerelease }}
       root-sbd: ${{ needs.defaults.outputs.root-sbd }}
     secrets: inherit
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -224,7 +207,7 @@ jobs:
     name: Notifier
     needs:
       - defaults  # so we can access `inputs.root-sbd` that could have defaulted to `inputs.model`
-      - version-tag  # implies all the spack.yaml-related checks have passed, has appropriate version for the prerelease build
+      - version  # implies all the spack.yaml-related checks have passed, has appropriate version for the prerelease build
```
aidanheerdegen, line 210: Comment message below still references `${{  needs.version-tag.outputs.release }}` and `${{ needs.version-tag.outputs.prerelease }}`.

.github/workflows/deploy-2-start.yml
```diff
@@ -64,11 +65,14 @@ jobs:
...
             ${{ secrets.HOST_DATA }}
 
-      - name: Prerelease Modulefile Name
+      - name: Prerelease spack.yaml Modifications
         if: inputs.type == 'prerelease'
         # Modifies the name of the prerelease modulefile to the
-        # `pr<number>-<commit>` style. For example, `access-om3/pr12-2`
+        # `pr<number>-<commit>` style. For example, `access-om3/pr12-2`.
+        # Also removes the `@git.VERSION` specifier for Prereleases so
+        # we don't have to shift tags around.
```
aidanheerdegen, line 73: What does it replace the `@git.VERSION` with?

aidanheerdegen (changes requested):
Questions questions...

.github/workflows/deploy-2-start.yml
CodeGat: With nothing, so it's just `- access-om2-bgc` for example. The version is implicit.

.github/workflows/ci.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/101/commits/adefb3953ea1ddd9291a51f0ebd7f08cecf8f6be

.github/workflows/deploy-2-start.yml
aidanheerdegen: So how do we make sure we're building the PR version?

CodeGat: The spack.yaml itself is the PR version - it is the file that is deployed

aidanheerdegen: Happy days!

aidanheerdegen (approved):
Nice work tracking down the error BTW.

🚀

PR #103, opened 2024-07-01
CD: Root SBD Defaults

### LINKED ISSUE #102: Metadata DB Upload: Model SBD not being interpreted
See the two spaces between `./metadata/output` and `mom5` [here](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/9675328146/job/26735226645#step:6:1) and the interpretation of `mom5` as the model name [here](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/9675328146/job/26735226645#step:6:24).

This is due to `${{ inputs.root-sbd }}` not interpreting the `default: ${{ inputs.model }}` [here](https://github.com/ACCESS-NRI/build-cd/blob/d7c57427766587787f130fe84ad455cde1752fcc/.github/workflows/deploy-2-start.yml#L33). There are no errors noting that this doesn't work, but [GitHubs own documentation](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_iddefault) notes that it must be a `string`. Indeed, we cannot have [dynamic defaults yet](https://github.com/orgs/community/discussions/25218). 

- [x] Update the workflow so the `inputs.root-sbd` has a sensible default.
- [x] Add to the `devdocs` this interesting fact

CodeGat:
This issue stems from this line:
https://github.com/ACCESS-NRI/build-cd/blob/d7c57427766587787f130fe84ad455cde1752fcc/.github/workflows/deploy-2-start.yml#L33
We cannot set dynamic defaults like this, only a basic string default. 

The CI pipeline already creates it's own `inputs.root-sbd` default (essentially `inputs.model-name`), as it is required for editing the modulefile name to be in `pr` format for Prereleases.
The CD pipeline also requires this, but we hadn't added it. Details on the underlying cause are in the linked issue. 

In this PR:
* Made `cd.yml` set a dynamic `inputs.root-sbd` default based on `inputs.model`
* Made all `inputs.root-sbd` required

 - [x] We will also have to delete and manually upload the `meta-db` entries that did not contain the correct `model` name. Currently, that seems to be all deployed models last month. 
 - [x] Will also need to link a `ACCESS-NRI/access-esm1.5` PR that adds the (now required) `inputs.root-sbd` argument. 

Closes #102

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,6 +10,13 @@ on:
         type: string
         required: true
         description: The model that is being tested and deployed
+      root-sbd:
+        type: string
+        required: false
+        # default: The ${{ inputs.model }} above
+        description: |
+          The name of the root Spack Bundle Definition, if it is different from the model name.
+          This is often a package named similarly in ACCESS-NRI/spack-packages.
   # Callers usually have the trigger:
   # push:
   #   branches:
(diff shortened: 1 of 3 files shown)


CodeGat:
@utkarshgupta95 would you be able to list the entries for the build provenance DB? I suspect that there will be a bunch of `mom5` models in there as a model rather than as a component.

CodeGat:
Linking https://github.com/ACCESS-NRI/ACCESS-ESM1.5/pull/7

aidanheerdegen (approved):
LGTM

PR #108, opened 2024-07-15
added model status field

utkarshgupta95:
- Adding status field to model
- A script to change the status of a model from spack_hash


CodeGat:
Linking https://github.com/ACCESS-NRI/experiment-metadb/issues/16

CodeGat (approved):
Looks good!

PR #113, opened 2024-08-02
added release_url to component build

### LINKED ISSUE #112: Add explicit fields in the table for the GitHub url for the model components
Currently, the frontend of the build provenance DB attempts to create the url of the version of the model components.

That doesn't work for model components that don't have a package name equal to the repository name (eg. `um7` and `um_v7`).

Add in a field for the model component for the url. Update the json parsing logic required. @CodeGat to update the workflow to get the url via `spack find` to turn into JSON.

utkarshgupta95:
This PR is the combined effort of @utkarshgupta95 adding a `release_url` to the DB schema for model components, and @CodeGat wrangling the model component repository url from `spack` and passing it into said script. 

In this PR:
* `tools/release_provenance`: Update schema to include new `release_url` field. Update test data.
* `deploy-2-start.yml`: Split Gadi deployment and Gadi metadata-gathering into separate steps. Get model component URLs from the `spack python` command and convert to a json file
* `generate-build-metadata.bash`: Read in above file and add it to the json model+component API that we have going here. 

## Limitations
* Each `spack python` command takes around 10s as it is concretizing the spec to get the package details. 

Resolves #112

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -114,20 +114,40 @@ jobs:
           spack env activate ${{ inputs.env-name }}
           spack --debug install --fresh ${{ vars.SPACK_INSTALL_PARALLEL_JOBS }} || exit $?
           spack module tcl refresh -y
+          EOT
 
-          # Obtain metadata
-          spack find --paths > ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}/spack.location
-          spack find --format '{hash} {prefix}' | jq --raw-input --null-input '[inputs | split(" ") | {(.[0]): (.[1])}] | add' > ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}/spack.location.json
+      - name: Get metadata from ${{ inputs.deployment-environment }}
+        env:
+          SPACK_ENV_PATH: ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}
+        run: |
+          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          spack find --paths > ${{ env.SPACK_ENV_PATH }}/spack.location
+          spack find --format '{hash} {prefix}' | jq --raw-input --null-input '[inputs | split(" ") | {(.[0]): (.[1])}] | add' > ${{ env.SPACK_ENV_PATH }}/spack.location.json
+
+          # Get the repos associated with the packages for the build database
+          jq -n '{}' > ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json
+          for pkg in ${{ vars.BUILD_DB_PACKAGES }}; do
+            # TODO: Is there a way to get the git attribute without concretizing?
+            pkg_repo=$(spack python -c "import spack.spec; print(spack.spec.Spec('$pkg').concretized().package.git)")
+            jq \
+              --arg p "$pkg" \
+              --arg r "$pkg_repo" \
+              '. += {($p): ($r)}' \
+              ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json > ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json.tmp
+            mv -f ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json.tmp ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json
+          done
 
           spack env deactivate
-          echo "$(date): Deployed ${{ inputs.model }} ${{ inputs.version }} with spack-packages ${{ steps.versions.outputs.packages }}, spack-config ${{ steps.versions.outputs.config }}" >> ${{ steps.path.outputs.root }}/release.log
           EOT
 
       # Release
       - name: Get Release Metadata
+        env:
+          # TODO: Can we put both envs above in a $GITHUB_ENV file instead?
+          SPACK_ENV_PATH: ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -114,20 +114,56 @@ jobs:
...
 
-          # Obtain metadata
-          spack find --paths > ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}/spack.location
-          spack find --format '{hash} {prefix}' | jq --raw-input --null-input '[inputs | split(" ") | {(.[0]): (.[1])}] | add' > ${{ steps.path.outputs.spack }}/var/spack/environments/${{ inputs.env-name }}/spack.location.json
+          spack find --paths > ${{ env.SPACK_ENV_PATH }}/spack.location
+          spack find --format '{hash} {prefix}' | jq --raw-input --null-input '[inputs | split(" ") | {(.[0]): (.[1])}] | add' > ${{ env.SPACK_ENV_PATH }}/spack.location.json
+
+          # Get the repos associated with the packages for the build database
+          jq -n '{}' > ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json
+          for pkg in ${{ vars.BUILD_DB_PACKAGES }}; do
```
aidanheerdegen, line 132: Where is `vars.BUILD_DB_PACKAGES` defined?

.github/workflows/deploy-2-start.yml
```diff
@@ -114,20 +114,56 @@ jobs:
...
+            # TODO: Is there a way to get the git attribute without concretizing?
```
aidanheerdegen, line 133: Isn't it already concretized in the environment where the model was built?

.github/workflows/deploy-2-start.yml
```diff
@@ -114,20 +114,56 @@ jobs:
...
+
+            # We get the version of $pkg from spack.lock, and then strip
+            # potential 'git.' and '=VERSION' parts from 'git.TAG=VERSION'
+            version=$(jq --compact-output --raw-output \
+              --arg p "$pkg" \
+              '.concrete_specs | to_entries[].value | select(.name == $p) | .version
+                | match("^(?:git.)?([^=]*)")
+                | .captures[0].string' \
+              ${{ env.SPACK_ENV_PATH}}/spack.lock
+            )
```
aidanheerdegen, line 144: It doesn't have to happen in this PR, but this really looks like something that would benefit from being a function we could write some tests for. At the very least it would then be explicit about what is, and what is not, supported.

For reviewing this it would be helpful to show some output of this running on the command line. Even paste in an example `spack.lock` file so a reviewer could easily run some test code.

.github/workflows/deploy-2-start.yml
```diff
@@ -114,20 +114,56 @@ jobs:
...
+            pkg_release_url="${pkg_repo_url%.*}/releases/tag/${version}"
```
aidanheerdegen, line 145: Would be good to have a comment that has an example `pkg_repo_url` because it's difficult from inspection to know what this is supposed to be doing. It's deleting everything after the first `.`, but without an example I'm not sure what is left.

.github/workflows/deploy-2-start.yml
```diff
@@ -114,20 +114,56 @@ jobs:
...
+
+            echo "$pkg pkg_repo_url is $pkg_repo_url, pkg_release_url is $pkg_release_url, version is $version"
+            jq \
+              --arg p "$pkg" \
+              --arg r "$pkg_release_url" \
+              '. += {($p): ($r)}' \
+              ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json > ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json.tmp
+            mv -f ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json.tmp ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json
```
aidanheerdegen, line 153: This is reformatting the `build-db-pkgs.json` file? Could you add a comment about why this is necessary?

aidanheerdegen (commented):
I have questions! :)

.github/workflows/deploy-2-start.yml
CodeGat: It is a variable defined in the calling model repository that is a space-separated list of packages to include as entries in the `component_build` section of the provenance DB. For example, `access-om2` is `mom5 cice5 oasis3-mct libaccessom2` I think.

.github/workflows/deploy-2-start.yml
CodeGat: It is, but I can't access existing concretized things from the python interpreter I don't think. 
It takes about 10 seconds to concretize per package...

.github/workflows/deploy-2-start.yml
aidanheerdegen: Tah.

.github/workflows/deploy-2-start.yml
CodeGat: So, we cant read from and redirect to the same `build-db-pkgs.json` (like `jq '...' build-db-pkgs.json > build-db-pkgs.json`) without potentially corrupting the file. So we read from `build-db-pkgs.json`, write to `build-db-pkgs.json.tmp` and the `mv` that file back to `build-db-pkgs.json`. See https://stackoverflow.com/a/42718624/16458915

.github/workflows/deploy-2-start.yml
CodeGat: The `spack.lock` I've been using for testing is the `access-esm1.5 2024.06.0` release. When I run the above I get the expected `access-esm1.5_2024.06.20`.

.github/workflows/deploy-2-start.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/113/commits/e653d8ec319fe3f2c4875f74c6b92a0c9ea7f812

aidanheerdegen (approved):
LGTM

PR #116, opened 2024-08-14
Import Commit Signing Keys for Workflow Commits

### LINKED ISSUE #115: Allow Workflow-Authoured Verified Commits
Currently, we use `access-bot` for our workflow-authored commits, but we don't use the associated signing key to make the commits show up as `verified`. 

See https://github.com/ACCESS-NRI/dev-docs/wiki/Continuous-integration-and-Deployment#verified-commits-from-a-workflow for the solution.

CodeGat:
In this PR:
* Import commit-signing key using org secrets across the pipeline
* Remove force pushes

Closes #115


jo-basevi (approved):
LGTM

PR #117, opened 2024-08-21
ci.yml: Update `MODULE_VER` creation logic to allow for multiple `/` in module name 

CodeGat:
See failed https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/10483277546/job/29035729585?pr=76

Given the update in https://github.com/ACCESS-NRI/schema/pull/34, we now allow more `/` in modulefile names. However, the check that the version of the package matches the version of the projection used a naive `split on /` method, which would not keep later `/`s. 

In this PR:
* Update the section getting the module version to use bash parameter splitting to split on the first `/` and keep the rest. 

References https://github.com/ACCESS-NRI/ACCESS-OM2/pull/76

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,8 @@ jobs:
               DEP_VER=$(yq ".spack.packages.\"$DEP\".require[0] | match(\"^@(?:git.)?([^=]*)\").captures[0].string" spack.yaml)
             fi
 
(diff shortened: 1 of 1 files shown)


dougiesquire (approved):
I don't know `yq` so I'm probably not the best person to review. However, I did run the modified lines on a `spack.yaml` using a `$DEP` with and without a `/`. Both gave the correct `MODULE_VER`

aidanheerdegen (approved):
LGTM

PR #119, opened 2024-08-28
Dynamic `spack` version in Deployment Comment

### LINKED ISSUE #110: Deployment Comment still uses hardcoded `0.21` spack version string
## Background
See https://github.com/ACCESS-NRI/build-cd/blob/db1ad07d516f3f9dbaa13fb3f73d62b6b1ac1057/.github/workflows/ci.yml#L233

This should be the actual version of `spack` deployed, from the `config/versions.json` file in the model repository. 

## Solution

Use the `check-config` job and run the `access-nri/build-cd/.github/actions/validate-repo-version` for `spack`. Note that the action will have to be modified as we refer to the `releases/<version>` branch of `spack` as just `version`. Furthermore, there are a few `${{ steps.version.outputs.packages }}` which don't make sense. Update to `${{ steps.jq.outputs.version }}`.

CodeGat:
We didn't have the infrastructure in place before to be able to get the spack version into the deployment comment - it used to have to be manually updated. This PR makes this not the case anymore. 

In this PR:

- **ci.yml: We now validate the spack version and pass it to the deployment comment**
- **validate-repo-version: Can now verify spack version existence, fixed incorrect templating**

Closes #110

### DIFF
--- a/.github/actions/validate-repo-version/action.yml
+++ b/.github/actions/validate-repo-version/action.yml
@@ -27,12 +27,8 @@ runs:
           exit 1
         fi
 
-        # We omit the 'releases/v' from the spack branch in versions.json, so we add it back here.
-        if [[ "${{ inputs.repo-to-check }}" == "spack" ]]; then
-          echo "version=releases/v${version}" >> $GITHUB_OUTPUT
-        else
-          echo "version=${version}" >> $GITHUB_OUTPUT
-        fi
+        echo "version=${version}" >> $GITHUB_OUTPUT
+        echo "spack-branch-version=releases/v${version}" >> $GITHUB_OUTPUT
 
     # Verify that the repository exists at the given ref
     - name: Version Check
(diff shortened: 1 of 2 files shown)


.github/actions/validate-repo-version/action.yml
```diff
@@ -23,27 +23,27 @@ runs:
...
-        ref: ${{ steps.versions.outputs.packages }}
+        # In the case where we are checking spack, we need to add the
+        # 'releases/v' to the ref name, obtained in the previous step
+        ref: ${{ inputs.repo-to-check == 'spack' && steps.jq.outputs.spack-branch-version || steps.jq.outputs.version }}
         path: repo
 
-    - name: Failure Notifier
-      if: failure()
+    - name: Version Check Failure Notifier
+      if: failure() && steps.check.outcome == 'failure'
```
jo-basevi, line 45: Is only `steps.check.outcome == 'failure'` needed here?

jo-basevi (approved):
Looks good to me

.github/actions/validate-repo-version/action.yml
CodeGat: Yeah, because we fail earlier in the workflow for the other fail conditions

PR #130, opened 2024-09-18
Rework `settings-2-deploy.yml` to Remove Local Reading of Files on Remote

### LINKED ISSUE #129: Deployment Settings Update Attempts to Access Local Runner File From Gadi
Currently, we attempt to read `config/settings.json` while `ssh`ed into a deployment target such as `Gadi`. See first and last lines linked: https://github.com/ACCESS-NRI/build-cd/blob/d7e95c55c13dfb4097bd2d379a0f685d4b8729b1/.github/workflows/settings-2-deploy.yml#L56-L65

This would mean that the Update step would fail as we can't access `${{ env.CONFIG_SETTINGS_PATH }}` (`./config/settings.json`) while `ssh`ed into Gadi.

CodeGat:
This PR fixes the issue of attempting to read runner-local `config/settings.json` on a remote target (such as `Gadi`). Whoops. 

In this PR:
* Move local reading of `config/settings.json` into earlier, pre-`ssh` connection step
* Remove `versions` output to a step since we don't really need it. 
* Use `while read ... done <<< "$newline_separated_things"` since it's super cool
* Quoted a variable in `if [ -n "$failed" ]` because `-n` doesn't work properly without quotes

Closes #129

### DIFF
--- a/.github/actions/validate-deployment-settings/action.yml
+++ b/.github/actions/validate-deployment-settings/action.yml
@@ -87,7 +87,7 @@ runs:
           fi
         done
 
-        if [ -n $failed ]; then
+        if [ -n "$failed" ]; then
           msg="Some commits referenced do not exist in access-nri/spack. Check the workflow logs."
           echo "::${{ inputs.error-level }}::$msg"
           echo "msg=$msg" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


jo-basevi (approved):
Looks good to me! It's neat there's only one `jq` query now. I didn't know about `while read ... done <<< ` so that was great to learn about!

CodeGat:
Yeah I know! it's a bit annoying having to have the thing we're iterating on at the bottom of the `while`, but what can ya do...

PR #136, opened 2024-09-24
Multiline GitHub Output Fix

### LINKED ISSUE #135: Error Setting Output in `settings-*` workflow
See failing run https://github.com/ACCESS-NRI/build-cd/actions/runs/11003760127/job/30553463542

Determine if the output is being set in a format that GitHub understands.

CodeGat:
There was an error processing multiline GitHub output, so we have to use a heredoc like in https://github.com/orgs/community/discussions/116619#discussioncomment-8994849. A simplified example of both is presented in this workflow: https://github.com/codegat-test-org/test/actions/runs/11004389439

In this PR:
- Put in missing ')' in `spack-config jq`
- Use a heredoc for multiline GITHUB_OUTPUT

Closes #135


jo-basevi (approved):
LGTM!

PR #139, opened 2024-09-24
Quote Hyphenated `spack-config` `jq` Property

CodeGat:
See failing run https://github.com/ACCESS-NRI/build-cd/actions/runs/11007422156/job/30563335709

This is due to not escaping a hyphenated property name like `spack-config` in `"\(.value.spack-config)"`. 

In this PR:
- Update usages of `spack-config` property to be escaped, like `"\(.value."spack-config")"`

### DIFF
--- a/.github/workflows/settings-2-deploy.yml
+++ b/.github/workflows/settings-2-deploy.yml
@@ -45,7 +45,7 @@ jobs:
           updates=$(jq --compact-output --raw-output \
             --arg env "${{ inputs.deployment-environment }}" \
             --arg type "${{ inputs.spack-type }}" \
-            '.deployment[$env][$type] | to_entries[] | "\(.key) \(.value.spack-config)"' \
+            '.deployment[$env][$type] | to_entries[] | "\(.key) \(.value."spack-config")"' \
             ${{ env.CONFIG_SETTINGS_PATH }}
           )
 


.github/workflows/settings-2-deploy.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           updates=$(jq --compact-output --raw-output \
             --arg env "${{ inputs.deployment-environment }}" \
             --arg type "${{ inputs.spack-type }}" \
-            '.deployment[$env][$type] | to_entries[] | "\(.key) \(.value.spack-config)"' \
+            '.deployment[$env][$type] | to_entries[] | "\(.key) \(.value."spack-config")"' \
```
aidanheerdegen, line 48: Does double nesting of `"` quotes work?

aidanheerdegen (commented):
A question.

.github/workflows/settings-2-deploy.yml
CodeGat: Yes, verified locally:
```bash
$ jq -cr --arg env "Gadi" --arg type "Prerelease" '.deployment[$env][$type] | to_entries[] | "\(.key) \(.value."spack-config")"' config/settings.json 
0.22 2024.07.05
```

aidanheerdegen: Wow. All other languages I've ever used would not like this at all. `jq` you crazy outlier. Never let anyone change you.

aidanheerdegen (approved):
I say let `jq` be `jq`.

CodeGat:
Possibly the `\(...)` variable substitution is a construct that ignores the outer quotation marks?

Issue #140, opened 2024-09-24
`settings.json`: `Release 0.22 spack` Does Not Contain Hash Due to `access-nri/spack` Rebase


CodeGat:
See warning in `Validate [config/settings.json] Deployment Settings` here: https://github.com/ACCESS-NRI/build-cd/actions/runs/11007422156/job/30563301689#step:5:245

This is due to ACCESS-NRI/spack@21da7d7e2b5e2680cd9d2e0a2fb4a7d13d8baa9d no longer existing on the `releases/v0.22` branch of `access-nri/spack`. This is because after syncing our fork with the upstream for its `0.22.2` updates, @harshula rebased our ACCESS-NRI-specific fixes on top of that. 

We need to come to a decision where either:

- We remove the check that verifies that the commit hash referenced is in the `release/v$VERSION` branch. This would remove a guardrail that prevents people from modifying the `0.XX` install to a commit hash not within that major version. Relevant code here: https://github.com/ACCESS-NRI/build-cd/blob/ab2b346275632e203996392442df7b628b94e3ec/.github/actions/validate-deployment-settings/action.yml#L83-L120
- We figure out a way to keep our forks `spack` fixes commits stable

aidanheerdegen:
We're at war with our smarter selves. I wonder who will win?

harshula:
Ironically, this is probably one of the arguments for using patch files instead of a branch that tracks upstream with local changes as commits.

harshula:
Hi @CodeGat , Could we do something like:
```
diff --git a/config/settings.json b/config/settings.json
index 94fe836..7c3fbf2 100644
--- a/config/settings.json
+++ b/config/settings.json
@@ -5,16 +5,19 @@checks
       "Release": {
         "0.20": {
           "spack": "6812713cf470b473a607f0de0e8e1cf53f804fb7",
+          "access-spack": "x",
           "spack-config": "2024.03.22"
         },
         "0.22": {
           "spack": "21da7d7e2b5e2680cd9d2e0a2fb4a7d13d8baa9d",
+          "access-spack": "y",
           "spack-config": "2024.07.05"
         }
       },
       "Prerelease": {
         "0.22": {
           "spack": "21da7d7e2b5e2680cd9d2e0a2fb4a7d13d8baa9d",
+          "access-spack": "z",
           "spack-config": "2024.07.05"
         }
       }
```
where:
`spack` denotes the latest upstream commit
[...]

PR #142, opened 2024-10-04
Disable 'Verify Spack Commits Exist' Check

CodeGat:
Disabling the `Verify Spack Commits Exist` Check as the unstable hashes on `access-nri/spack` make it so the check will almost always fail. See the linked issue for [the long term solution](https://github.com/ACCESS-NRI/build-cd/issues/140#issuecomment-2392638311). 

References #140


harshula (approved):
Thanks!

PR #148, opened 2024-10-21
config/settings.json: update to use an external Python

harshula:
* Using an external Python should reduce the inode usage.


CodeGat (approved):
Looks good to me :)

PR #161, opened 2024-11-12
`config/settings.json` Pipeline Fixes

### LINKED ISSUE #160: Linking Workflow Run to Closed PR Fails, Hash-Equivalent Tags Not Showing As Unchanged
See https://github.com/ACCESS-NRI/build-cd/actions/runs/11791433825/job/32843449208 for the failing link to PR. 
Furthermore, see the annotations as part of https://github.com/ACCESS-NRI/build-cd/actions/runs/11791433825 to note that the `spack-config` repos are "changing" from a given hash to the equivalent tag. 

Checkout the appropriate repo so the PR Linking works, and find a way to compare hashes and tags effectively.

CodeGat:
This PR fixes a few small bugs, each confined to their own file. 

Namely:
* The `spack-config` tag check was just listing the tags but not actually checking for their existence in the `spack-config` repo. This PR actually checks each tag. Fixed in https://github.com/ACCESS-NRI/build-cd/pull/161/commits/9563eaf9ee3fdfb68a1c2b538a8dac66123bcc97
* When updating the version of `spack-config`, we were comparing the current `HEAD` hash against the git tag name, rather than the hash associated with the git tag. This would lead to the annotation erroneously stating that the `spack-config` version was always "changing" when in actual fact we were checking out the same tag that we were on. Fixed in https://github.com/ACCESS-NRI/build-cd/pull/161/commits/f4fd81c37f2b49f35c1a08cd85094d0daab5ca4c
* We attempt to link the deployment of settings to the newly-closed PR, but we hadn't checked out a repo before running `gh`. Whoops! Fixed in https://github.com/ACCESS-NRI/build-cd/pull/161/commits/3efad6bb8c273be3504e93f1d46d50eb7b2ffdb2

In this PR:

- **validate-deployment-settings: loop around tags in settings.json, update condition**
- **settings-2-deploy: Convert spack-config tag into commit for version comparison**
- **settings-1-update.yml: Checkout repo before using `gh`**

Closes #160

### DIFF
--- a/.github/workflows/settings-2-deploy.yml
+++ b/.github/workflows/settings-2-deploy.yml
@@ -100,7 +100,8 @@ jobs:
           set +e
           while read -ra update; do
             version=${update[0]}
-            new_commit=${update[1]}
+            new_tag=${update[1]}
+
             current_head_commit=$(git -C ${{ secrets.SPACK_INSTALLS_ROOT_LOCATION }}/$version/spack-config rev-parse HEAD)
 
             if [ $? -eq 128 ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/settings-2-deploy.yml
```diff
@@ -110,8 +111,11 @@ jobs:
 
             git -C ${{ secrets.SPACK_INSTALLS_ROOT_LOCATION }}/$version/spack-config fetch
 
+            # Parse the above spack-config tag as a commit hash, for easy comparison. The '^{}' syntax gives the hash of the tagged commit, not the hash of the tag itself
```
aidanheerdegen, line 114: Deep in the git-fu you are. 

Skilled you have become.

aidanheerdegen (approved):
TMLGID

PR #162, opened 2024-11-13
Get git URL of package without concretizing

### LINKED ISSUE #147: Find a Different Way to Get the Git URL of a Package Without Concretizing
See the step in https://github.com/ACCESS-NRI/ACCESS-ESM1.5/actions/runs/11396654421/job/31710870230#step:9:13 - overall this section takes 14 minutes to concretize the packages in `vars.BUILD_DB_PACKAGES`, which is almost as long as it took to build the package itself. 

See https://github.com/ACCESS-NRI/build-cd/blob/edfd0a8d4318b64c70c35537076f39c263c00bd8/.github/workflows/deploy-2-start.yml#L131

Determine if this change should be reverted or if there is a better one.

CodeGat:
In this PR:
* Instead of freshly concretizing the packages in `vars.BUILD_DB_PACKAGES` to get the git URL, use the `package_class` attribute. This means that instead of waiting 14 minutes for reconcretization of some packages, it will be almost instant. 

Closes #147

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -127,8 +127,7 @@ jobs:
           # Get the repos associated with the packages for the build database
           jq -n '{}' > ${{ env.SPACK_ENV_PATH }}/build-db-pkgs.json
           for pkg in ${{ vars.BUILD_DB_PACKAGES }}; do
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
*mashing approve button*

PR #166, opened 2024-11-14
Use `git reset` rather than `git checkout` for `spack-packages` update

### LINKED ISSUE #165: Rebased `spack-packages` on Remote Leads to Divergent History on Gadi
See run https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/11829981295/job/32962759210#step:8:29 - there is history divergence because of rebases on `spack-packages` `development` branches. 

Instead of using `git checkout --force $BRANCH`, which still has the divergent branch issue, use `git reset --hard origin/$BRANCH`, since we only care about what is done on the remote.

CodeGat:
In this PR:
* Replace `git checkout --force` with `git reset --hard origin/$BRANCH` so the deployment targets `spack-packages` always has the remotes history, rather than diverging. 

Closes #165

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -101,6 +101,7 @@ jobs:
 
           # Update spack-packages
           git -C ${{ steps.path.outputs.spack-packages }} fetch
+          git -C ${{ steps.path.outputs.spack-packages }} checkout --force ${{ steps.versions.outputs.packages }}
           git -C ${{ steps.path.outputs.spack-packages }} reset --hard origin/${{ steps.versions.outputs.packages }}
 
           # Enable spack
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -101,7 +101,7 @@ jobs:
 
           # Update spack-packages
           git -C ${{ steps.path.outputs.spack-packages }} fetch
-          git -C ${{ steps.path.outputs.spack-packages }} checkout --force ${{ steps.versions.outputs.packages }}
+          git -C ${{ steps.path.outputs.spack-packages }} reset --hard origin/${{ steps.versions.outputs.packages }}
```
harshula, line 104: Have you already switched to the branch? The steps I do are:
1) Fetch
2) Be in the correct branch
3) Sync current branch to origin's branch

CodeGat: Good point. For some reason I thought `git reset` just moved the `HEAD`, not the local branch. Fixed in https://github.com/ACCESS-NRI/build-cd/pull/166/commits/c3b20321f2dadbdaf052836db58387cd568bdb82

harshula (approved):
:crossed_fingers:

PR #172, opened 2024-11-21
Update PR Description with Prerelease Link

### LINKED ISSUE #152: Give current pre-release in the PR description
Currently, when a pre-release is built, the action opens a new comment with the instructions to find the pre-release.

One downside of that approach is comments for previous pre-releases stay visible although the information is out-dated. This can lead to people trying to access an non-existent pre-release and get confused.

It might be better to give the information directly at the bottom of the PR description. It is easier to find and should allow to remove information to old pre-releases and only keep the current, existing release.

This is done to some extent for readthedocs previews in this action: https://github.com/readthedocs/actions

Here is what it looks like for readthedocs:
Screenshot 2024-10-24 at 12 29 11 PM

CodeGat:
## Background

In this PR:
* Add a new step to the `notifier` job that updates the PR description with the latest attempted Prerelease. It'll look something like this:

----

🚀 The latest prerelease `access-om2/pr86-12` is [here](https://github.com/ACCESS-NRI/ACCESS-OM2/pull/86#issuecomment-2490129613) 🚀

----

## Testing

Tested `awk` and `gh` commands locally. 
Tested the entire step in my own org, with both creation and update steps: https://github.com/codegat-test-org/test/actions/runs/11962762802/job/33351840627 for PR https://github.com/codegat-test-org/test/pull/22

Closes #152

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,21 +285,18 @@ jobs:
       - name: PR Description Notifier
         env:
           GH_TOKEN: ${{ github.token }}
-          PR_BODY_FILE_LOCATION: ./body.txt
-          LATEST_PRERELEASE_SECTION_REGEX: "^:rocket: .* :rocket:$"
-          # We can't use many punctuation marks in the Prerelease Section because `sed` treats
-          # the replacement string as a regex string - no special regex chars without escaping
-          LATEST_PRERELEASE_SECTION: ":rocket: The latest prerelease `${{ needs.defaults.outputs.root-sbd }}/${{ needs.check-spack-yaml.outputs.prerelease }}` is here: ${{ steps.comment.outputs.comment-link }} :rocket:"
+          PR_BODY_PATH: ./body.txt
+          PR_BODY_PATH_UPDATED: ./updated.body.txt
+          PRERELEASE_SECTION_REGEX: "^:rocket: .* :rocket:$"
+          PRERELEASE_SECTION: ":rocket: The latest prerelease `${{ needs.defaults.outputs.root-sbd }}/${{ needs.check-spack-yaml.outputs.prerelease }}` is here: ${{ steps.comment.outputs.comment-link }} :rocket:"
         run: |
-          gh pr view ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --json body --jq .body > ${{ env.PR_BODY_FILE_LOCATION }}
+          gh pr view ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --json body --jq .body > ${{ env.PR_BODY_PATH }}
 
-          if grep -q '${{ env.LATEST_PRERELEASE_SECTION_REGEX }}' ${{ env.PR_BODY_FILE_LOCATION }}; then  # there is an existing prerelease section
-            # Replace the existing prerelease section - we use '|' as a delimiter because there are '/' in the replacement string
-            sed -i 's|${{ env.LATEST_PRERELEASE_SECTION_REGEX }}|${{ env.LATEST_PRERELEASE_SECTION }}|g' ${{ env.PR_BODY_FILE_LOCATION }}
+          if grep -q '${{ env.PRERELEASE_SECTION_REGEX }}' ${{ env.PR_BODY_PATH }}; then  # there is an existing prerelease section
+            # Replace the existing prerelease section
+            awk '{gsub(/${{ env.PRERELEASE_SECTION_REGEX }}/, "${{ env.PRERELEASE_SECTION }}")}1' ${{ env.PR_BODY_PATH }} > ${{ env.PR_BODY_PATH_UPDATED }}
           else  # a new section is added
-            echo -e '\n' >> ${{ env.PR_BODY_FILE_LOCATION }}
-            echo '---' >> ${{ env.PR_BODY_FILE_LOCATION }}
-            echo '${{ env.LATEST_PRERELEASE_SECTION }}' >> ${{ env.PR_BODY_FILE_LOCATION }}
+            awk '1; END {print "\n---\n${{ env.PRERELEASE_SECTION }}"}' ${{ env.PR_BODY_PATH }} > ${{ env.PR_BODY_PATH_UPDATED }}
           fi
 
-          gh pr edit ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --body-file ${{ env.PR_BODY_FILE_LOCATION }}
+          gh pr edit ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --body-file ${{ env.PR_BODY_PATH }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,7 +247,9 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - uses: access-nri/actions/.github/actions/pr-comment@main
+      - name: PR Comment Notifier
+        id: comment
+        uses: access-nri/actions/.github/actions/pr-comment@main
         with:
           comment: |
             :rocket: Deploying ${{ inputs.model }} `${{ needs.check-spack-yaml.outputs.release }}` as prerelease `${{ needs.check-spack-yaml.outputs.prerelease }}` with commit ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -279,3 +281,25 @@ jobs:
...
             If this is not what was expected, commit changes to `config/versions.json`.
             </details>
+
+      - name: PR Description Notifier
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY_FILE_LOCATION: ./body.txt
+          LATEST_PRERELEASE_SECTION_REGEX: "^:rocket: .* :rocket:$"
+          # We can't use many punctuation marks in the Prerelease Section because `sed` treats
+          # the replacement string as a regex string - no special regex chars without escaping
```
atteggiani, line 291: What do you mean by punctuation marks?

I think as much as trying to avoid some characters can be safe, it might not be safe enough depending on the values of `${{ needs.defaults.outputs.root-sbd }}`, `${{ needs.check-spack-yaml.outputs.prerelease }}` and `${{ steps.comment.outputs.comment-link }}` (if any of that contains metacharacters, especially `&`, it might be a problem).
A safer solution might be escaping `sed`'s replacement string before using it within the `sed` command. 
In case you want to be safer, [this answer](https://stackoverflow.com/a/29613573/21024780) has extensive solutions for escaping sed strings.

atteggiani (approved):
Looks good! 

I just added a few minor suggestions.

.github/workflows/ci.yml
```diff
@@ -279,3 +281,25 @@ jobs:
...
+        run: |
+          gh pr view ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --json body --jq .body > ${{ env.PR_BODY_FILE_LOCATION }}
+
+          if grep -q '${{ env.LATEST_PRERELEASE_SECTION_REGEX }}' ${{ env.PR_BODY_FILE_LOCATION }}; then  # there is an existing prerelease section
+            # Replace the existing prerelease section - we use '|' as a delimiter because there are '/' in the replacement string
+            sed -i 's|${{ env.LATEST_PRERELEASE_SECTION_REGEX }}|${{ env.LATEST_PRERELEASE_SECTION }}|g' ${{ env.PR_BODY_FILE_LOCATION }}
+          else  # a new section is added
+            echo -e '\n' >> ${{ env.PR_BODY_FILE_LOCATION }}
+            echo '---' >> ${{ env.PR_BODY_FILE_LOCATION }}
+            echo '${{ env.LATEST_PRERELEASE_SECTION }}' >> ${{ env.PR_BODY_FILE_LOCATION }}
```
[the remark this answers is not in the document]
CodeGat: For some reason I couldn't get the `EOF` to work properly...see https://github.com/codegat-test-org/test/actions/runs/11947542405/job/33303660670#step:2:8

.github/workflows/ci.yml
CodeGat: Yeah, fair enough. I just didn't want to have to do an escaped and non-escaped version of the `echo` variant :D I'll think up a solution...

atteggiani: I think you could have something like:
```bash
LATEST_PRERELEASE_SECTION='${{ env.LATEST_PRERELEASE_SECTION }}'
ESCAPED_LATEST_PRERELEASE_SECTION= ... # command to escape the string
```
and then for your `sed` command use:
```bash
sed -i 's|${{ env.LATEST_PRERELEASE_SECTION_REGEX }}|`"$ESCAPED_LATEST_PRERELEASE_SECTION|g" ${{ env.PR_BODY_FILE_LOCATION }}
```
and keep `${{ env.LATEST_PRERELEASE_SECTION }}` (or you can also use `$LATEST_PRERELEASE_SECTION`) for the `echo` command.

Note you might also want to use `shell: bash +H {0}` to avoid problems with the `!` character (history expansion)

.github/workflows/ci.yml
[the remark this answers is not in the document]
atteggiani: Hmm maybe it expects the end-of-file (`EOL`) to be right at the beginning of the line. 
Have you tried with the ["indented" heredoc](https://unix.stackexchange.com/a/76483)?
Otherwise it's ok with multiple echo :D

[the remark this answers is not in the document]
CodeGat: The "must use tabs" bit is a bit of a dealbreaker - vscode converts tabs to spaces and I don't want a section that only uses tabs :D

.github/workflows/ci.yml
CodeGat: I've managed to get around the escaping issue by using `awk` instead of `sed` - done in commit https://github.com/ACCESS-NRI/build-cd/pull/172/commits/3e65b4bfc90f16ab3b113eddbc9fa9ab436cd846

See the test runs for https://github.com/codegat-test-org/test/pull/22 in https://github.com/codegat-test-org/test/actions/runs/11964783047 and https://github.com/codegat-test-org/test/actions/runs/11964805872

.github/workflows/ci.yml
```diff
@@ -279,3 +281,22 @@ jobs:
...
+          PR_BODY_PATH: ./body.txt
+          PR_BODY_PATH_UPDATED: ./updated.body.txt
+          PRERELEASE_SECTION_REGEX: "^:rocket: .* :rocket:$"
+          PRERELEASE_SECTION: ":rocket: The latest prerelease `${{ needs.defaults.outputs.root-sbd }}/${{ needs.check-spack-yaml.outputs.prerelease }}` is here: ${{ steps.comment.outputs.comment-link }} :rocket:"
+        run: |
+          gh pr view ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --json body --jq .body > ${{ env.PR_BODY_PATH }}
+
+          if grep -q '${{ env.PRERELEASE_SECTION_REGEX }}' ${{ env.PR_BODY_PATH }}; then  # there is an existing prerelease section
+            # Replace the existing prerelease section
+            awk '{gsub(/${{ env.PRERELEASE_SECTION_REGEX }}/, "${{ env.PRERELEASE_SECTION }}")}1' ${{ env.PR_BODY_PATH }} > ${{ env.PR_BODY_PATH_UPDATED }}
```
atteggiani, line 297: This solution with `awk` seems a bit more cryptic to me, but it works.

It also requires you to have another file (even though I think you could output the result to the same `${{ env.PR_BODY_PATH }}` file), and it doesn't solve all issues with metacharacters etc., but with the current structure of `PRERELEASE_SECTION` should work fine.

Is the `PRERELEASE_SECTION` structure supposed to stay the same? Because in that case I wonder if it might be better to only substitute the needed portions (which I believe are `${{ needs.defaults.outputs.root-sbd }}/${{ needs.check-spack-yaml.outputs.prerelease }}` and `${{ steps.comment.outputs.comment-link }}`).

Anyway it's a matter of taste, I think there shouldn't be real issues with the current implementation.

atteggiani (dismissed):
Again, looks good but minor comments for the new implementation.

.github/workflows/ci.yml
CodeGat: Yes `awk` is a bit more cryptic, but the substitution string crucially is not a regex string (like in `sed`), just a regular one. Hence we could put whatever we wanted into it, metacharacters included. 
When I was looking into it, I hadn't seen anyone write to the file they were reading, only writing to a temp file and then `mv`ing it. 
The `PRERELEASE_SECTION` may change, who knows - having it in one place (the `env` var) is preferable to updating it in multiple.

.github/workflows/ci.yml
```diff
@@ -279,3 +281,22 @@ jobs:
...
+          else  # a new section is added
+            awk '1; END {print "\n---\n${{ env.PRERELEASE_SECTION }}"}' ${{ env.PR_BODY_PATH }} > ${{ env.PR_BODY_PATH_UPDATED }}
```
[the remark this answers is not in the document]
CodeGat: Haha, unfortunately having the `-e` flag means that echo will interpret the "`MODEL/VERSION`" section and fail. It'll have to be the triple `echo`. As a compromise, I'll better comment what it is doing.

.github/workflows/ci.yml
atteggiani: Makes sense.

.github/workflows/ci.yml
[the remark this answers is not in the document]
atteggiani: ooh right! 🙈 
Ok I give up ahah
Commenting seems a good compromise!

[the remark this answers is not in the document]
CodeGat: Commented on `awk` in https://github.com/ACCESS-NRI/build-cd/pull/172/commits/398a93384c76bc996a4abff121d44b5979c2d2e7 - what do you think?

atteggiani (approved):
LGTM!

PR #176, opened 2024-11-25
Add `!redeploy` Comment Command

### LINKED ISSUE #146: Deployment Comment `spack` Environment Name Not Filled In
See https://github.com/ACCESS-NRI/ACCESS-ESM1.5/pull/14#issuecomment-2421178883, specifically in the `Details and usage instructions` section. 

This line:

> This Prerelease is also accessible on Gadi via `/g/data/vk83/prerelease/apps/spack/0.22/spack` in the `access-esm1p5-` environment.

should be:

> This Prerelease is also accessible on Gadi via `/g/data/vk83/prerelease/apps/spack/0.22/spack` in the `access-esm1p5-pr14-3` environment.

This is probably due to referencing an incorrect template variable.

### LINKED ISSUE #173: Create a Comment Command That Redeploys the Current Commit
## Background

A common use case seems to be requiring a rebuild of the `HEAD` Prerelease when a model component using a `@git.BRANCH` reference is updated. The `@git.BRANCH` reference is better than using a `@git.TAG`/`@git.COMMIT` reference, as the others would either require adding a temporary tag, or copying the commit hash. It's also preferable to adding a comment to the `spack.yaml` noting the update, because that is an update that would need to be updated. 

Add a comment command that redeploys the current `HEAD`.

> [!NOTE]
> ~As with other ChatOps methods, this will not prevent/allow merging, unless there is a way to append checks to a given commit. Specifically, if one redeployed with a model component change that broke the model, **it would still be allowed to merge**. Although, https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status looks good...~ This is not the case anymore, adding checks from non-PR sources works fine, so we can prevent merging with `myrotvorets/set-commit-status-action`. 

## Workaround

* Add an empty commit, in which the message can note what was changed: `git commit --allow-empty -m "Redeploying to test out new component changes"`
* Comment in the `spack.yaml` the hash used in the components `@git.BRANCH HEAD` 
* Use the `@git.TAG` reference

CodeGat:
## Background

See https://github.com/ACCESS-NRI/model-deployment-template/pull/9#discussion_r1853234344 - instructions for people to add empty commits to redeploy their model is a bit fiddly. So, create a ChatOps command to redeploy, instead. 

The majority of this PR is updating the `ci.yml` job to handle an explicit `inputs.pr` rather than inferring from `github.event.pull_request`/`github.[head|base]_ref`, since we now offer two entrypoints into `ci.yml` - `on.pull_request`, and `on.issue_comment` (specifically, commenting `!redeploy`). 

We also have `redeploy-pre`/`redeploy-post` jobs added when the trigger is a `!redeploy` comment, in which we set the HEAD commit status to pending, run the rest of the `ci.yml`, then update the HEAD commit again based on the prerelease deployment status.

In this PR:
- `ci.yml`
  - Updated deployment comment prerelease version templating, references #146
  - Added new required input `pr`, set pr metadata in `defaults`  and clone all repos at `inputs.pr`, replace all `github.*` contexts with `inputs.pr`/`needs.default.outputs.*`
  - Add `redeploy-pre` and `redeploy-post` jobs that handle commit status updates on `!redeploy`
- `validate-repo-version` action: Added new required input `pr`, updated references in `ci.yml`

## Requirements for Model Deployment Repositories

Before merge, there must be updates to deployment repositories like so:

```diff
 jobs:
   pr-ci:
     name: CI
-    if: github.event_name == 'pull_request' && github.event.action != 'closed'
+    if: >-
+      (github.event_name == 'pull_request' && github.event.action != 'closed') ||
+      (github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '!redeploy'))
     uses: access-nri/build-cd/.github/workflows/ci.yml@main
     with:
       model: ${{ vars.NAME }}
       # root-sbd: if different from vars.NAME
+      pr: ${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.event.issue.number }}
     permissions:
       pull-requests: write
       contents: write
+      statuses: write
     secrets: inherit
 
   pr-comment:
     name: Comment
-    if: github.event_name == 'issue_comment'
+    if: github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '!bump') 
     uses: access-nri/build-cd/.github/workflows/ci-comment.yml@main
     with:
       model: ${{ vars.NAME }}
       # root-sbd: if different from vars.NAME
```

## Testing

- The majority of testing was done in this PR: https://github.com/ACCESS-NRI/ACCESS-TEST/pull/7 - note the successes at the end!
- All `gh` commands were tested locally

-----

Closes #173
Closes #146

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,8 @@ jobs:
         id: commenter
         uses: access-nri/actions/.github/actions/commenter-permission-check@main
         with:
+          # This means that commenters who use `!redeploy` must have at least `write` perms
+          # in the repository. `write` is probably the best fit. 
           minimum-permission: write
 
       - name: React to Comment
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,12 +111,14 @@ jobs:
         uses: access-nri/build-cd/.github/actions/validate-repo-version@main
         with:
           repo-to-check: spack-packages
+          pr: ${{ inputs.pr }}
 
       - name: Validate spack version
         id: spack
         uses: access-nri/build-cd/.github/actions/validate-repo-version@main
         with:
           repo-to-check: spack
+          pr: ${{ inputs.pr }}
 
       - name: Checkout build-cd Config
         uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ name: CI
 run-name: ${{ inputs.model }} CI
 # NOTE: This workflow requires:
 # permissions.pull-requests:write
-# permissions.contents:write
 # secrets:inherit with an appropriate GitHub Environment for deployment in the caller
 on:
   workflow_call:
@@ -30,7 +29,10 @@ on:
   #   paths:
   #     - config/**
   #     - spack.yaml
-
+  # issue_comment:
+  #   types:
+  #     - created
+  #     - edited
 env:
   SPACK_YAML_MODEL_YQ: .spack.specs[0]
 jobs:
@@ -265,8 +267,8 @@ jobs:
   # -----------------------------
   prerelease-deploy:
     name: Deploy to Prerelease
-    # This will create a `spack` environment with the name `<model>-pr<pull request number>-<commit number>`.
-    # For example, `access-om3-pr13-3` for the deployment of access-om3 based on the third commit on the PR#13.
+    # This will create a `spack` environment with the name `<model>-pr<pull request number>-<deployment number>`.
+    # For example, `access-om3-pr13-3` for the third deployment on the PR#13.
     needs:
       - defaults  # so we can access `inputs.root-sbd` that could have defaulted to `inputs.model`
       - check-spack-yaml  # implies all the spack.yaml-related checks have passed, has appropriate version for the prerelease build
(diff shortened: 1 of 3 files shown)


CodeGat:
Note, this will have to be reworked as we are hitting the limit of 5 nested workflow calls.

tmcadam:
@CodeGat Will this be used on public repos? If so, would it be an issue with anyone being able to add the magic comment to PRs?

CodeGat:
Good point @tmcadam - I'll add a similar step that checks the permissions of the user.

tmcadam:
@CodeGat Could this also be achieved with `workflow_dispatch`, I think you are able to choose a branch(?). But maybe that is the branch for the action rather than the code that is checked out.

CodeGat:
Indeed, it's for the branch of the action. And I think it's less overhead to do a command in the PR rather than a workflow dispatch where they need to remember the PR number

.github/workflows/ci.yml
```diff
@@ -127,11 +127,23 @@ jobs:
       # String to describe the overall check
       commit-status-description: ${{ steps.commit-status-args.outputs.description }}
     steps:
+      - name: Check commenter permissions
+        id: commenter
+        uses: access-nri/actions/.github/actions/commenter-permission-check@main
+        with:
+          minimum-permission: write
```
tmcadam, line 134: Could we add a comment to here to make sure future devs know that it's important not change this without understanding what it does

tmcadam (changes requested):
Looks good, just one comment

.github/workflows/ci.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/176/commits/4aca73c6712dbaba291cdc09ee9029bcadb60d5c

tmcadam (approved):

aidanheerdegen (approved):
Looks great, thanks!

Even though this has added a lot of code, the refactor makes it much cleaner I think.

I also appreciated the very thorough overview in the top level post. The context helped a lot with understanding what was being done, which made it a lot quicker for me to review.

PR #177, opened 2024-11-26
Return All Deployment Results

CodeGat:
See equal deployment numbers https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/10#issuecomment-2499625946 and https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/10#issuecomment-2499635474. 

This is due to the default pagination in `gh api` calls to be 30 per page, and only the first page. So as deployments are made in this branch, some are evicted to the second page. 

This PR adds the `--paginate` flag to the `gh api` call, which attempts to return all the results.


harshula (approved):
:crossed_fingers:

Issue #180, opened 2024-12-02
spack.yaml: allow the specs section to contain variants


harshula:
Due to `build-cd` not allowing variants in the specs section, this type of hack is required: https://github.com/ACCESS-NRI/spack-packages/commit/84a010b752d07fbbf8536b63a8ac4b11ec61a26f in the SBRs. i.e. A separate branch had to be created to allow building the variant.

CodeGat:
Closing this as the PRs linked are closed

PR #181, opened 2024-12-02
Update Projection Check to Allow Variants

CodeGat:
## Background

In this PR:

* Update the model projection version check so it can ignore later, optional, space-separated variants in the `spack.specs` string. We use a regex capture group because they can do more than a `split` function. 

For example, given the `spack.yaml`s `spack.spec[0]` being `MODEL@git.VERSION type=thingo +variant1 ~debug` or `MODEL@git.VERSION`, both versions will be `VERSION`. 

## Testing

To test this regex, I validated it using regex101: https://regex101.com/r/cbQMyG/1.
I also validated it against ACCESS-OM2's `spack.yaml`. 

References #180


harshula (approved):
:crossed_fingers:

PR #188, opened 2024-12-10
Deployment Number Pagination Fix

### LINKED ISSUE #187: Issue with Deployment Number with `--paginate`d results
See nonexistent deployment number with https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/12249294255/job/34170365865#step:4:22

This is due to `gh`s `--paginate` results, which put multiple "pages" in their own root `[]`, which leads to [the jq](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/12249294255/job/34170365865#step:4:6) eventually ending up with at least two results, like `16 0` in which there are `16` deployments in the first page, and `0` in the second. 

`jq --slurp` the result from `gh api` instead.

CodeGat:
## Background

Due to `gh api --paginate` returning multiple arrays of pages of results for much-deployed-to repositories, we need a way to add the pages of deployments together to get correct results. 
First, we `--slurp` the results (put each 'page' into an array), then we call `jq` explicitly (since we can't use the `--jq` flag with `--slurp` for some reason), and then do a much-the-same filter (but one level deeper) to find the number of deployments to a given branch. 

Furthermore, we also do a quick fix to the way the deployment numbers are calculated. Because the number of `$pr_deployment` events only account for previous deployments (not the currently deploying one!) and the number of `$comment_deployments` account for all previous AND the current one, we only need to add one to the next deployment number if the next deployment is a traditional, pr-based one. Capiche?

In this PR:
* We fix the `$deployment_number` when the `gh api` results are paginated. 
* We also fix an issue with the deployment numbers incrementing incorrectly, see example: https://github.com/ACCESS-NRI/ACCESS-TEST/pull/10#issuecomment-2530528914

## Testing 

Done in https://github.com/ACCESS-NRI/ACCESS-TEST/pull/10 - note the multiple successful deployment numbers

Closes #187

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,10 @@ jobs:
             --json comments \
             --jq '[.comments[] | select(.body | startswith("!redeploy"))] | length'
           )
-          next_deployment_number=$((pr_deployments + comment_deployments + 1))
+          # Since the number of $pr_deployments do not include the current deployment (yet),
+          # but $comment_deployments do, we need to increment the next deployment number by one if it is a pr deployment.
+          next_deployment_is_pr_deployment=${{ github.event_name == 'pull_request' && '1' || '0' }}
+          next_deployment_number=$((pr_deployments + comment_deployments + next_deployment_is_pr_deployment))
           echo "Next Deployment Number is $next_deployment_number"
           echo "next-deployment-number=$next_deployment_number" >> $GITHUB_OUTPUT
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
           # but $comment_deployments do, we need to increment the next deployment number by one if it is a pr deployment.
           next_deployment_is_pr_deployment=${{ github.event_name == 'pull_request' && '1' || '0' }}
           next_deployment_number=$((pr_deployments + comment_deployments + next_deployment_is_pr_deployment))
-          echo "Next Deployment Number is $next_deployment_number"
+          echo "Next Deployment Number is $pr_deployments + $comment_deployments + $next_deployment_is_pr_deployment = $next_deployment_number"
           echo "next-deployment-number=$next_deployment_number" >> $GITHUB_OUTPUT
 
   redeploy-pre:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -98,18 +98,22 @@ jobs:
...
+          # We --slurp the results because --paginate introduces potentially multiple array results
           pr_deployments=$(gh api \
-            -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" \
-            --paginate \
-            /repos/${{ github.repository }}/deployments \
-            --jq '[.[] | select(.ref == "${{ steps.pr.outputs.head }}")] | length'
+              -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" \
+              --paginate --slurp \
+              /repos/${{ github.repository }}/deployments \
+            | jq '[select(.[][].ref == "${{ steps.pr.outputs.head }}")] | length'
```
aidanheerdegen, line 106: I got an error running this on the command line
```bash
gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" --paginate /repos/access-nri/access-test/deployments | jq 'select(.[][].ref = "paginate_deployments") | length'
```
error:
```
jq: error (at <stdin>:0): Cannot index string with string "ref"
```
But this works:
```bash
gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" --paginate /repos/access-nri/access-test/deployments | jq '[select(.[].ref = "paginate_deployments")] | length'
```

aidanheerdegen (commented):
I have a question.

.github/workflows/ci.yml
CodeGat: You need to add the `--slurp` option to the first command for it to work.

aidanheerdegen (approved):
LGTM!

.github/workflows/ci.yml
CodeGat: Time for an update, then! :laughing:

PR #189, opened 2024-12-11
Add workaround to `spack` fetch issue

CodeGat:
## Background

Since the `spack` bug with failing to fetch package repositories might be a way off, automate the existing workaround in the linked issue.

To do this, we simply run `git fetch` over all the source repos that spack maintains, before we do the actual deployment step. We are hoping that this step will be deleted once the underlying bug is fixed. 

In this PR:
* Add step before deployment that fetches all repositories

## Testing

Testing of the script was run on a copy of `Gadi Prerelease` that had failing builds due to this bug, and it rectified the issue. 

References #2

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -100,12 +100,10 @@ jobs:
             exit 0
           fi
 
-          cd "${{ steps.path.outputs.root }}/git_repos" || exit 1
-          # ls might not be space-separated all the time, xargs space-separates
-          for package_repo in $(ls | xargs); do
-            git -C "./$package_repo" remote show origin
-            git -C "./$package_repo" fetch --verbose --update-head-ok origin +refs/heads/*:refs/heads/*
-            echo "--------------------------"
+          cd "${{ steps.path.outputs.root }}/git_repos" || exit 0
+
+          for package_repo in *; do
+            git -C "./$package_repo" fetch --verbose --update-head-ok origin "+refs/heads/*:refs/heads/*"
           done
           EOT
 
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -89,6 +89,26 @@ jobs:
             spack.yaml \
             ${{ secrets.USER }}@${{ secrets.HOST_DATA }}:${{ vars.SPACK_YAML_LOCATION }}/${{ inputs.model }}.spack.yaml
 
+      - name: Fetch git sources for packages to build
+        # Workaround for ACCESS-NRI/spack#2.
+        # TODO: Remove once underlying spack issue is fixed
+        run: |
+          ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+
+          if [ ! -d "${{ steps.path.outputs.root }}/git_repos" ]; then
+            echo "::warning::Unable to find git_repos folder under ${{ steps.path.outputs.root }} - spack will attempt fetch instead"
+            exit 0
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -89,6 +89,26 @@ jobs:
...
+        # TODO: Remove once underlying spack issue is fixed
+        run: |
+          ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+
+          if [ ! -d "${{ steps.path.outputs.root }}/git_repos" ]; then
+            echo "::warning::Unable to find git_repos folder under ${{ steps.path.outputs.root }} - spack will attempt fetch instead"
+            exit 0
+          fi
+
+          cd "${{ steps.path.outputs.root }}/git_repos" || exit 1
```
aidanheerdegen, line 103: This exit condition won't b0rk the whole workflow though right?

.github/workflows/deploy-2-start.yml
```diff
@@ -89,6 +89,26 @@ jobs:
...
+          # ls might not be space-separated all the time, xargs space-separates
+          for package_repo in $(ls | xargs); do
+            git -C "./$package_repo" remote show origin
+            git -C "./$package_repo" fetch --verbose --update-head-ok origin +refs/heads/*:refs/heads/*
```
aidanheerdegen, line 107: I am nervous and think surely this *must* need to be quoted ... but I guess not?

.github/workflows/deploy-2-start.yml
```diff
@@ -89,6 +89,26 @@ jobs:
...
+          ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+
+          if [ ! -d "${{ steps.path.outputs.root }}/git_repos" ]; then
+            echo "::warning::Unable to find git_repos folder under ${{ steps.path.outputs.root }} - spack will attempt fetch instead"
+            exit 0
+          fi
+
+          cd "${{ steps.path.outputs.root }}/git_repos" || exit 1
+          # ls might not be space-separated all the time, xargs space-separates
+          for package_repo in $(ls | xargs); do
```
aidanheerdegen, line 105: I think 
```bash
for package_repo in *; do
```
should work too.

aidanheerdegen (approved):
I have questions as per .. but happy enough.

.github/workflows/deploy-2-start.yml
CodeGat: It will, actually. But I'll update it so it shouldn't. Technically that condition shouldn't even happen, since we checked above that the directory exists.

.github/workflows/deploy-2-start.yml
CodeGat: Indeed it does! Will update it

.github/workflows/deploy-2-start.yml
CodeGat: It hasn't given any problems so far, but I'll make it safer.

.github/workflows/deploy-2-start.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/189/commits/bd10d5511dd208955cff0d9a5ad12557c5e5cb44

.github/workflows/deploy-2-start.yml
CodeGat: Quoted it and it works just as fine - in https://github.com/ACCESS-NRI/build-cd/pull/189/commits/bd10d5511dd208955cff0d9a5ad12557c5e5cb44

aidanheerdegen (approved):
LGTM. Thanks!

PR #191, opened 2024-12-11
Only do Deployment Version Checks if PR is not a Draft

### LINKED ISSUE #190: Don't require a bumped version if it is a drafted PR
It seems that more and more, people are using the Prerelease area not as a stepping stone to a fully-fledged release, but as a testing ground that doesn't have any possibility of merging. 

We want to be a bit kinder in this case, so we should disable the check on overall versions if the PR is in a drafted state.

CodeGat:
In this PR:
* Add a fairly basic condition that we should only bother with checking that the overall model version is bumped if it is not a draft. 

Note that there is a small hole with this approach - if a PR is drafted, but undrafted after the final commit before a merge, the model version check will not have fired, and the PR could be erroneously merged with the same version. Luckily, the Release CD will pick up on this and fail, which means one just has to reopen the PR, and reset the `main` branch. If a better solution is sought, I can create an issue. 

Closes #190

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,6 +255,13 @@ jobs:
           data-location: ./spack.yaml
 
       - name: Check Model Version Modified
+        # We don't want to fire off model deployment version checks if the PR is never going to be merged.
+        # We determine this by checking if either the pull request, or the pull request that a comment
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM

Issue #109, opened 2024-07-26
`ci-comment`: Failed to checkout using `secrets.GH_COMMIT_CHECK_TOKEN`


CodeGat:
## Background

See error comment: https://github.com/ACCESS-NRI/ACCESS-OM2/pull/76#issuecomment-2251720536

The `!bump major` comment command doesn't work, as [we cannot checkout our repo](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/10103695385/job/27941489639#step:2:21) using the `secrets.GH_COMMIT_CHECK_TOKEN`. 
The token was expired, so it was regenerated. Even that failed. Investigate this. 

Also, this workflow was absent from the fix to `inputs.root-sbd.default: ${{ inputs.model }}` - [see here](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/10103695385/job/27941489639#step:2:18). We can't set dynamic inputs, so add a job that sets the `inputs.root-sbd`  as a default to `env.root-sbd`.

## Workaround

Current workaround is to manually update the version of the model, rather than using `!bump [major|minor]`

Pinging @dougiesquire as he found this error - Thanks!

PR #192, opened 2024-12-12
`!bump` comment command fix

CodeGat:
The majority of this issue comes down to a token mismatch (which isn't visible in this PR), and the fact that I'd missed a place where I'd set `root-sbd.default: inputs.model`, which is invalid in GitHub Actions. 

Furthermore, I'd failed to do `secrets.inherit` in the calling workflows, which is why this is a major update. 

## Testing

Testing took place in https://github.com/ACCESS-NRI/ACCESS-TEST/pull/11 - note the successful bump, as well as the deployment created after the bump!

In this PR:
- **ci-comment.yml: Set default root-sbd, update permissions**
- **ci-comment.yml: Made env vars set after defaults job completes**

References #109

### DIFF
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -2,6 +2,7 @@ name: Comment Command
 run-name: ${{ inputs.model }} Comment Command
 # NOTE: This workflow requires:
 # permissions.contents:write
+# permissions.pull-requests:write
 on:
   workflow_call:
     inputs:
@@ -12,7 +13,7 @@ on:
       root-sbd:
         type: string
         required: false
-        default: ${{ inputs.model }}
+        # default: ${{ inputs.model }}
         description: |
           The name of the root Spack Bundle Definition, if it is different from the model name.
           This is often a package named similarly in ACCESS-NRI/spack-packages.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-comment.yml
```diff
@@ -12,7 +13,7 @@ on:
       root-sbd:
         type: string
         required: false
-        default: ${{ inputs.model }}
+        # default: ${{ inputs.model }}
```
aidanheerdegen, line 16: Should this just be deleted? Or a comment added saying this isn't possible, so workflow below effectively does this?

aidanheerdegen (dismissed):
LGTM. Thanks for fixing this.

.github/workflows/ci-comment.yml
CodeGat: Good point. Will update.

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/192/commits/d6c41a87cc69ea1f63e5b7303a714d4602aa85e0

CodeGat:
I shall ask for forgiveness since the change is minor since your last approval - just some comments!

PR #198, opened 2024-12-13
ci.yml: Updated release version to ignore variants in `.spack.specs[0]`

### LINKED ISSUE #185: Versioning is erroneously taking optional `~variants` in `.spack.specs[0]` as a Version
See "Deploying..." part of https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/15#issuecomment-2526546678

Due to #180, there was a bit of an oversight. We take `.spack.specs[0]` as the version for the comment (and maybe even the CD?). But now we can add `variants`, we need to update the places that use the above as a version.

CodeGat:
In this PR:
* Make so the release version ignores optional variants after the version string. Uses a capture function rather than split. 

Tested locally on all deployment repositories - none failed. 

Closes #185


jo-basevi (approved):
LGTM

Issue #199, opened 2024-12-17
Fix Up Multi-Target TODOs


CodeGat:
There are some clearly marked sections in `build-cd` that will fail if we end up having multiple targets. Fix these, and generify some of the comments that make reference to Gadi specifically.

CodeGat:
# Old Comment

:rocket: Deploying access-esm1.6 `dev_2024.12.0` as prerelease `pr27-5` with commit fa3a855ee802a431cf635c73459ca524f2d7b406

<details>
<summary>Details and usage instructions</summary>

This `access-esm1.6` model will be deployed as:
* `dev_2024.12.0` as a Release (when merged).
* `pr27-5` as a Prerelease (during this PR).

This Prerelease is accessible on Gadi using:
```bash
module use /g/data/vk83/prerelease/modules
module load access-esm1p6/pr27-5
```
where the binaries shall be on your `$PATH`.
This Prerelease is also accessible on Gadi via `/g/data/vk83/prerelease/apps/spack/0.22/spack` in the `access-esm1p6-pr27-5` environment.
</details>

:hammer_and_wrench: Using: spack `0.22`, spack-packages `2024.12.0`, spack-config `2024.11.27`
<details>
<summary>Details</summary>

It will be deployed using:
* `access-nri/spack` on branch [`0.22`](https://github.com/ACCESS-NRI/spack/tree/releases/v0.22)
* `access-nri/spack-packages` version [`2024.12.0`](https://github.com/ACCESS-NRI/spack-packages/releases/tag/2024.12.0)
* `access-nri/spack-config` version [`2024.11.27`](https://github.com/ACCESS-NRI/spack-config/releases/tag/2024.11.27)

If this is not what was expected, commit changes to `config/versions.json`.
</details>

CodeGat:
# New Comment 1 (Close to original)

:rocket: Deploying `access-esm1.6`  Prerelease `pr27-5` with commit access-nri/access-esm1.6@fa3a855ee802a431cf635c73459ca524f2d7b406

<details>
<summary>Details and usage instructions</summary>

When using the below modules, the binaries shall be on your `$PATH`. 

-------

This `access-esm1.6` model will be deployed to `Gadi` as:
* `dev_2024.12.0` as a Release (when merged).
* `pr27-5` as a Prerelease (during this PR).

This Prerelease is accessible on `Gadi` using:
```bash
module use /g/data/vk83/prerelease/modules
module load access-esm1p6/pr27-5
[...]

Issue #194, opened 2024-12-12
Update Deployment Comment to Include Link to Deployment Run


CodeGat:
This would be especially useful for the deployments that aren't easily visible (e.g., the `!redeploy` ones)

PR #203, opened 2024-12-23
Move jobs around so they can be part of a single parallel workflow

CodeGat:
## Background
Although preliminary work had been done on allowing multiple HPC targets (notably by [creating a config file](https://github.com/ACCESS-NRI/build-cd/blob/main/config/settings.json) that allows for multiple HPCs, using GitHub Environments that correspond to multiple HPCs, and having inputs to existing workflows take a variable GitHub Environment), we aren't fully there yet. 

The major change is that we start the matrixing(?) of the overall pipeline much, much earlier. Each matrix job contains the `check-spack-yaml`, `check-config` and `deploy` jobs, and the rest of the pre- and post- matrix jobs are relatively unchanged. Although, the deployment comment that is generated is now no longer relatively static, but must be generated on the fly depending on how many deployment targets there are. Yuck. 

### Annoying Workaround Diversion...

One major annoyance is that GitHub does not allow dynamic outputs from matrix jobs to be consolidated easily. For example, say we have a matrix job (with `id` `get-location`) where each instance of the job returns an output (called `location`). If a later job `needs` `get-location`, the `needs.get-location.outputs.location` will be the value of the _last completed successful_ instance of the matrix job (since the output `location` is continuously overwritten), which is both not what we want, AND nondeterministic. 

In order to get all the outputs of the matrix job we need to do a workaround in which we write the "outputs" as an artifact of an instance of the matrix job, and then read those artifacts in later jobs. See https://github.com/ACCESS-NRI/build-cd/pull/203/files#diff-5e593466d0b143a180dbe89a953c8d5b832f52f3610e393f4f7542d5e7f4fbf9R54

## How It Works, Visually

Chicken-scratches on how the workflow looked after and before are below:

jobs_moved

## Testing

Testing was done primarily in https://github.com/ACCESS-NRI/ACCESS-TEST/pull/15 (starts working after [this event](https://github.com/ACCESS-NRI/ACCESS-TEST/pull/15#event-15742531759)). Note that we only deploy to `Gadi`, but it should work just as well with multiple targets. 

References #199

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -35,6 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       root-sbd: ${{ steps.root-sbd.outputs.default }}
+      targets: ${{ steps.target.outputs.matrix }}
     steps:
       - name: root-sbd
         id: root-sbd
@@ -45,9 +46,27 @@ jobs:
             echo "default=${{ inputs.root-sbd }}" >> $GITHUB_OUTPUT
           fi
 
+      - name: Get deployment settings.json
+        uses: actions/checkout@v4
+        with:
+          repository: access-nri/build-cd
+          ref: main
+
+      - name: Generate Deployment Target Matrix
+        id: target
+        run: |
+          targets=$(jq --compact-output '.deployments' config/settings.json)
+          echo "$targets"
+          echo "targets=$targets" >> $GITHUB_OUTPUT
+
   verify-settings:
     name: Verify Deployment Settings
     runs-on: ubuntu-latest
+    needs:
+      - defaults
+    strategy:
+      matrix:
+        target: ${{ fromJson(needs.defaults.outputs.targets) }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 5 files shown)


aidanheerdegen:
> Note that we only deploy to Gadi, but it should work just as well with multiple targets.

Would there be benefit in creating another (test) deployment target so the parallelisation can be tested?

If so, a few options spring to mind:

1. Deploy to a `test` directory in another project, e.g. `tm70`
2. Deploy to an ephemeral GitHub runner (is that possible?)
3. Create a test instance on Nectar that we can use to deploy to

CodeGat:
Good point! I'll create a new test spack instance on Gadi and give it a go!

CodeGat:
Tested it with another spack instance on `Gadi` (called `GadiTest`) and it worked fine: https://github.com/ACCESS-NRI/ACCESS-TEST/pull/15#issuecomment-2570048216

aidanheerdegen (approved):
LGTM! Awesome!

CodeGat:
Moved into dev branch referred to in https://github.com/ACCESS-NRI/build-cd/issue/121

CodeGat:
This solves https://github.com/ACCESS-NRI/build-cd/issues/194 as well

PR #210, opened 2025-01-17
Replace `env.SPACK_YAML_MODEL_YQ` to accomodate for multi-target-format `spack.yaml`

### LINKED ISSUE #208: Support Multi-target `spack.yaml`
Requires https://github.com/ACCESS-NRI/model-deployment-template/issues/15

## Background

Assuming that we are going with https://github.com/ACCESS-NRI/model-deployment-template/issues/15#issuecomment-2574265200 (but still allowing the existing `spack.specs[0]` method). 

This will require modification to a bunch of the `env.SPACK_YAML_MODEL_YQ` (as well as other `yq` stuff). 
It also might be a good time to implement functionality noted in https://github.com/ACCESS-NRI/build-cd/issues/183, but if that isn't possible, one could use the alternative operator `//` on the existing `yq` filters.

CodeGat:
Closes #208

## Background

Since we are (pre-emptively) going with the multi-target-compliant `spack.yaml` referenced in https://github.com/ACCESS-NRI/model-deployment-template/issues/15#issuecomment-2574265200 (https://github.com/ACCESS-NRI/model-deployment-template/issues/15), we'll need to update the `yq` filter to accomodate for root packages that are stored in `.spack.definitions[].root_package[]` rather than `.spack.specs[0]`. 

In this PR:
* Update `env.SPACK_YAML_MODEL_YQ` across the pipeline to first check for the new multi-target-compliant type of definition, and if that's `null`, go back to the traditional `.spack.specs[0]`. 

## Testing

Tested `yq` filter locally using https://github.com/ACCESS-NRI/ACCESS-OM2/blob/d907f3314a9956875baaaaf2b4d7b6be6fa81926/spack.yaml and https://github.com/ACCESS-NRI/ACCESS-OM2/blob/69168932f1b3414a5c5850a8c62063d0c7b48628/spack.yaml

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -27,7 +27,9 @@ on:
   #     - spack.yaml
 
 env:
-  SPACK_YAML_MODEL_YQ: .spack.specs[0]
+  # This attempts to get the spack.yaml model first via the multi-target
+  # .spack.definitions.root_package method, then the traditional '.spack.specs[0]'.
+  SPACK_YAML_MODEL_YQ: (.spack.definitions[] | select(."root_package") | .[][]) // .spack.specs[0]
   METADATA_PATH: /opt/metadata
   OUTPUTS_PATH: /opt/outputs
 jobs:
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -13,7 +13,7 @@ on:
       root-sbd:
         type: string
         required: false
-        # The equivalent default is set in the defaults job below. 
+        # The equivalent default is set in the defaults job below.
         # default: ${{ inputs.model }}
         description: |
           The name of the root Spack Bundle Definition, if it is different from the model name.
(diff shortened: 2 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -27,7 +27,9 @@ on:
   #     - spack.yaml
 
 env:
-  SPACK_YAML_MODEL_YQ: .spack.specs[0]
+  # This attempts to get the spack.yaml model first via the multi-target
+  # .spack.definitions.root_package method, then the traditional '.spack.specs[0]'.
+  SPACK_YAML_MODEL_YQ: (.spack.definitions[] | select(."ROOT_PACKAGE") | .[][]) // .spack.specs[0]
```
jo-basevi, line 32: Should it be `root_package` as I think it is case sensitive?

jo-basevi: Oops nevermind - I was just looking at the files used in the testing section but I saw there is a commit for upper case ROOT_PACKAGE and that's matches the schema

jo-basevi (approved):
Looks good

.github/workflows/cd.yml
CodeGat: Yeah, it's the uppercase one this time...just to disambiguate it from other `spack.definitions[]` that users might add - `ROOT_PACKAGE` and `ROOT_SPEC` are required!

Issue #209, opened 2025-01-17
Support Choice In Deployment Targets


CodeGat:
Rather than forcing users to deploy to multiple deployment targets when they may only want to deploy to one, allow them to choose via an updated `config/versions.json`. 

Something like:
```json
{
  "Gadi": {
    "spack": "0.22",
    "spack-packages": "2024.11.23"
  },
  "Setonix": {
    "spack": "0.23",
    "spack-packages": "2025.01.01"
  }
}
```
To deploy to `Gadi` and `Setonix`. If one only wanted to deploy to `Gadi`, the `Setonix` object can be removed.

CodeGat:
However, this same file is used to deploy releases. If one removed the `Setonix` key, we could create Release deployments for a single target, which would conflict with https://github.com/ACCESS-NRI/build-cd/issues/200#issuecomment-2572320835. This would not work, so I shall close this issue.

CodeGat:
In future we could have the solution in the description, but it would not affect what is deployed (the `vars` is the point of truth)

CodeGat:
Closing as referenced PRs are merged

PR #211, opened 2025-01-21
Support Deployment Target Choice For Deployment Repositories

CodeGat:
References #209

> [!IMPORTANT]
> This updated changes the model-deployment repositories `ci.yml`s `pr-closed` job to accepting a `root-sbd` argument rather than a `model` one. This is a major update, which will be incorporated into the overall `dev-121-multi-target-workflows`. 

## Background

Instead of the deployment targets coming from `build-cd` and being unchangeable from the model deployment repositories perspective, allow choice based on model deployment repository by picking deployment targets, set as repo-level `vars`. Specifically, this PR adds `vars.RELEASE_DEPLOYMENT_TARGETS` and `vars.PRERELEASE_DEPLOYMENT_TARGETS` to all model deployment repositories. These *MUST* be set for any repositories implementing the future `v4` of the infrastructure that this PR is a part of. 

In this PR:
* Updated target matrix generation to take into account model deployment repository preferences for deployment targets, set as a space-separated list in the above `vars` - turned this into a custom action (`get-target-matrix`)
* In order for the action to be usable by `ci-closed.yml`, I needed to rework some of the logic. This led to a big simplification in that pipeline. 

## Testing

The `jq` steps were tested locally on different `$valid_targets`, and `$repo_release_targets` - especially the intersection in `$targets`.


jo-basevi (dismissed):
Looks great!

Feel free to ignore this but as the logic is similar in both files, it almost could be worth a reusable action with an input to pass `${{ vars.PRERELEASE_DEPLOYMENT_TARGETS }}` or `${{ vars.RELEASE_DEPLOYMENT_TARGETS }}`

CodeGat:
Thats a good idea. Made even good-er-er by the fact that I forgot to do similar logic in `ci-closed.yml`! I'll do that update now.

CodeGat:
Thanks for your feedback @jo-basevi - I've created the action in https://github.com/ACCESS-NRI/build-cd/pull/211/commits/895fdd033c076956527897b9e091883d9f7464d1 and added it to `ci.yml` and `cd.yml` in https://github.com/ACCESS-NRI/build-cd/pull/211/commits/cd03bf13ba92764b33ee207ac2173ae06d64dfe2.

As I'd noted before, I forgot to do `ci-closed.yml` in a similar way. I'd updated that workflow, and then realised that we don't actually need to break it up into `ci-closed.yml` -> `undeploy-1-setup.yml` -> `undeploy-2-start.yml` - the middle workflow can be incorporated into the first. So I'd done a bit of rejigging of that workflow in https://github.com/ACCESS-NRI/build-cd/pull/211/commits/4d8a7c8262e8744bb62c79387ad14ab2e322a791

.github/actions/get-target-matrix/action.yml
```diff
@@ -0,0 +1,44 @@
...
+      id: generate
+      # These are used in the deploy job to determine the targets to deploy to - irrespective of the type of deployment.
+      # We use the list of valid targets (from build-cd) intersected with targets set in the caller repository to determine the final list.
+      run: |
+        valid_targets=$(jq --raw-output --compact-output \
+          '.deployment | keys' \
+          config/settings.json
+        )
+        repo_prerelease_targets=$(jq --null-input --raw-output --compact-output \
+          --arg valids "${{ inputs.deployment-targets }}" \
```
jo-basevi, line 33: Is it `${{ inputs.targets }}` rather than `${{ inputs.deployment-targets }}`? 
Maybe it makes more sense to have a more general variable name `repo_targets` rather than `repo_prerelease_targets`?

CodeGat: Ahh, forgot to generify that bit! Thanks. Done in https://github.com/ACCESS-NRI/build-cd/pull/211/commits/a106dabf266cf89415a7559f68a6fdd4dbb423af

CodeGat:
Also validated the action works with https://github.com/codegat-test-org/test/actions/runs/12921504911/job/36035654828#step:2:1

jo-basevi (approved):
Looks good - Nice change on merging undeploy into one workflow

.github/workflows/cd.yml
```diff
@@ -50,18 +50,11 @@ jobs:
...
-          repository: access-nri/build-cd
-          ref: main
-
       - name: Generate Deployment Target Matrix
         id: target
-        run: |
-          targets=$(jq --compact-output --raw-output '.deployment | keys' config/settings.json)
-          echo "$targets"
-          echo "targets=$targets" >> $GITHUB_OUTPUT
+        uses: access-nri/build-cd/.github/actions/get-target-matrix@v3
```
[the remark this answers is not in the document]
CodeGat: To `v4`, I'm just looking to change all of them in a separate PR.

PR #214, opened 2025-01-22
Interim `model-db` Update: Only Ingest `Gadi` Metadata

### LINKED ISSUE #212: Interim Solution: Update existing Model DB Infra to only use `Gadi`
Since it will be a significant amount of work, and we don't anticipate many people will require a proper Release deployment to any target other than `Gadi` for a while, we will instead have an interim solution in which we still use the `Gadi` deployment details to update the current `model-db`. Prereleases don't update the DB, so they aren't affected. 

References #201 (the not-yet-planned multi-target model-db stuff)
References ACCESS-NRI/tracking-services#23 (the new DB app)

CodeGat:
Closes #212

## Background

Currently, the `model-db` section of the multi-target workflow doesn't work, because we renamed a bunch of the metadata/outputs to have target-specific filenames (e.g `Gadi.spack.lock`, `deploy-outputs.Gadi`). 

In this PR, we are restoring the original functionality in which the `model-db` is updated with `Gadi`-specific information, until we have an updated provenance database, which will come about through ACCESS-NRI/tracking-services#23.

In this PR:

- generate-build-metadata.bash: Reference explicit Gadi.* files
- tests/scripts/generate-build-metadata: Updated test data inputs/outputs
- cd.yml: Use Gadi-specific outputs/metadata for build-db upload

## Testing

Tested using the updated `tests/scripts/generate-build-metadata` tests.


jo-basevi (approved):
Looks good to me

PR #231, opened 2025-02-21
Update trigger: on push to default branch, and always on pull request

### LINKED ISSUE #229: `settings-*.yml`: Update branch protection to Default Branch rather than `main`
`main` doesn't really exist in `build-cd` - we use `vX` branches as the default branch. Use conditional execution rather than `on.<pull_request|push>.branches` since we can't know the default branch then. 

Update `on.pull_request.branches` to all, and do conditional execution of the first job for $default_branch

CodeGat:
Closes #229

## Background

Since we are moving from a `main` default branch to a `vX` branch (where `X` is the latest major version of `build-cd`), we are updating the trigger for this job to:

* The default branch `on.push`
* Any branch `on.pull_request`

Provided it updates the `config/settings.json` file. 

## The PR

In this PR:

* Update trigger condition for running settings checks and deployments.

### DIFF
--- a/.github/workflows/settings-1-update.yml
+++ b/.github/workflows/settings-1-update.yml
@@ -1,13 +1,13 @@
 name: Config Settings Update
 on:
   push:
-    branches:
-      - main
+    # Below condition is enforced by the `if` condition in 
+    # the `setup-settings-update` job:
+    # branches:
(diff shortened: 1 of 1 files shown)


harshula (approved):
:crossed_fingers:

PR #218, opened 2025-01-23
Multi-HPC Workflows

CodeGat:
Closes #121 super-issue
Close https://github.com/ACCESS-NRI/build-cd/issues/213
Close https://github.com/ACCESS-NRI/build-cd/issues/196

> [!IMPORTANT]
> This is a major update to build infrastructure. Therefore, there are changes required to Model Deployment Repositories before the `v4` branch can be used. These are noted in the 'Major Model Deployment Repository Changes' section

## Background

This super Pull Request contains a rework of `build-cd` infrastructure to support deployment to multiple HPCs at once. The major changes are noted below:

* **`spack.yaml` Updates**: `spack.yaml` file now allows for variations in compiler/target/variants based on HPC target via mutually-exclusive `spack.definitions[].when` clauses, [example here](https://github.com/ACCESS-NRI/ACCESS-OM2/blob/ce5fdd27badabcbff464d50fcdc4df8e80a74123/spack.yaml).  Schema defined in [`1-0-4.json`](https://github.com/ACCESS-NRI/schema/blob/main/au.org.access-nri/model/spack/environment/deployment/1-0-4.json), implementation done in https://github.com/ACCESS-NRI/build-cd/issues/208 and discussed in https://github.com/ACCESS-NRI/model-deployment-template/issues/15

* **Model Deployment Repository Choice In Deployment Targets**: Since not all model deployment repositories may want to deploy to all valid deployment targets (namely, all the things defined in `config/settings.json`), model deployment repositories can set the repo-level `vars.[PRE]RELEASE_DEPLOYMENT_TARGETS` to choose Prerelease/Release targets. This is most useful for future Releases where we don't want to yet deploy officially to other HPC targets. Done in https://github.com/ACCESS-NRI/build-cd/pull/211

* **Release Notes and Deployment Comment Updates**:  Examples of the [former](https://github.com/ACCESS-NRI/build-cd/issues/199#issuecomment-2555915869) and [latter](https://github.com/ACCESS-NRI/build-cd/issues/200#issuecomment-2574326112), the final update from https://github.com/ACCESS-NRI/build-cd/pull/207

* **Releases**: A notable decision: a single deployment version encompasses multiple deployment targets, see https://github.com/ACCESS-NRI/build-cd/issues/200#issuecomment-2572320835

[...]

PR #232, opened 2025-02-21
Update Settings Matrix Generation to Adhere to `settings.schema.json`

CodeGat:
## Background

Fixes regression introduced in #218 in which we attempted to generate a matrix of targets without adhering to the `config/settings.schema.json`, leading to an empty (null) matrix and a failed workflow: see https://github.com/ACCESS-NRI/build-cd/actions/runs/13451909066/job/37587737547?pr=228
## The PR

In this PR:
- Update matrix generation to conform with existing schema - the key is `deployment` not `deployments`, and we only require the keys, not the entire object. 

## Testing

Tested `jq` locally - ran `jq -cr '.deployment | keys' config/settings.json` at a949ae057e07f7f9c07f58990ac1fb1e081a38ac and got expected `["Gadi"]`.

Tested matrix generated successfully in test workflow: https://github.com/codegat-test-org/test/actions/runs/13452297829

### DIFF
--- a/.github/workflows/settings-1-update.yml
+++ b/.github/workflows/settings-1-update.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Generate Deployment Target Matrix
         id: target
         run: |
-          targets=$(jq --compact-output '.deployments' config/settings.json)
+          targets=$(jq --compact-output '.deployment | keys' config/settings.json)
           echo "$targets"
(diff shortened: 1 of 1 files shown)


CodeGat:
Going to merge and ask for forgiveness on this one. Testing above is sufficient to ensure correctness.

PR #235, opened 2025-03-02
Allow `@latest`

### LINKED ISSUE #234: Allow `@latest` in spack manifest speclist
Since we tell users making local builds to use `@latest`, we should support them moving those `spack.yaml`s over into Prerelease, which means allowing `@latest` in both the schema and workflow.

CodeGat:
References #234

## Background

Local `spack` builds using a `spack.yaml` need to use `@latest` because they can't reference a not-yet-existing `@git.TAG` like prerelease builds can. To allow continuity between these types, allow `@latest` in the schema. 

## The PR

In this PR:

- **get-spack-root-spec action: Allow `@latest` as ref**
- **cd.yml: Add a check so that a `@latest` version is not deployed as a Release**

## Testing

Tested locally via `yq` on a `spack.yaml` of the form:

```yaml
spack:
  specs:
    - access-om2@latest
```

by running:

```bash
docker run -i --rm mikefarah/yq '.spack.specs[0] | capture("(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")' < spack.yaml
```

### DIFF
--- a/.github/actions/get-spack-root-spec/action.yml
+++ b/.github/actions/get-spack-root-spec/action.yml
@@ -65,9 +65,9 @@ runs:
 
         # Example of groups from the above:
         # name: anything before `@git.`. Ex: access-om2
-        # ref: anything after `@git.` but before an `=` (for =VERSION syntax) or before a space, `+` or `~` (for variants). Ex: 2025.01.0
+        # ref: anything after `@` or `@git.`; but before `=` (for =VERSION syntax) or ` `/`+`/`~` (for variants). Ex: 2025.01.0
         # version: anything after a `=`, but before a space, `+` or `~` (for variants). Ex: release
-        groups=$(yq '${{ steps.yq.outputs.filter }} | capture("(?<name>.+)@git\\.(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")' ${{ inputs.spack-manifest-path }})
+        groups=$(yq '${{ steps.yq.outputs.filter }} | capture("(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")' ${{ inputs.spack-manifest-path }})
 
         # Pull values from groups above
         name=$(yq '.name' <<< "$groups")
--- a/.github/actions/get-spack-root-spec/action.yml
+++ b/.github/actions/get-spack-root-spec/action.yml
@@ -68,9 +68,9 @@ runs:
         # ref: anything after `@` or `@git.`; but before `=` (for =VERSION syntax) or ` `/`+`/`~` (for variants). Ex: 2025.01.0
         # version: anything after a `=`, but before ` `/`+`/`~` (for variants). Ex: release
         
-        groups_regex="${{ steps.yq.outputs.filter }} | capture(\"(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*\")"
+        groups_regex='(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*'
 
-        groups=$(yq "$groups_regex" ${{ inputs.spack-manifest-path }})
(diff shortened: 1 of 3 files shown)


.github/actions/get-spack-root-spec/action.yml
```diff
@@ -65,9 +65,9 @@ runs:
 
         # Example of groups from the above:
         # name: anything before `@git.`. Ex: access-om2
-        # ref: anything after `@git.` but before an `=` (for =VERSION syntax) or before a space, `+` or `~` (for variants). Ex: 2025.01.0
+        # ref: anything after `@` or `@git.`; but before `=` (for =VERSION syntax) or ` `/`+`/`~` (for variants). Ex: 2025.01.0
         # version: anything after a `=`, but before a space, `+` or `~` (for variants). Ex: release
-        groups=$(yq '${{ steps.yq.outputs.filter }} | capture("(?<name>.+)@git\\.(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")' ${{ inputs.spack-manifest-path }})
+        groups=$(yq '${{ steps.yq.outputs.filter }} | capture("(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")' ${{ inputs.spack-manifest-path }})
```
atteggiani: An option to make this line a bit clearer could be to set the regexp in a different variable above this line and use it here.

atteggiani (dismissed):
Looks good to me.
I added a very minor comment, but feel free to ignore it.

.github/actions/get-spack-root-spec/action.yml
CodeGat: Not a bad idea, I'll do that!

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/235/commits/fabf572bac257c5120817b68514b1088e95e12ca, what do you think @atteggiani

atteggiani: I think you don't want bash expansion within the `group_regex`, so I would wrap it in single quotes instead of double:
```diff
- groups_regex="${{ steps.yq.outputs.filter }} | capture(\"(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*\")"
+ groups_regex='${{ steps.yq.outputs.filter }} | capture("(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*")'
```

Also, I was thinking more about only taking out the actual regex, without the filter, something like:
```bash
groups_regex='(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*'
groups=$(yq '${{ steps.yq.outputs.filter }} | capture("'"$groups_regex"'")')
```

CodeGat: Ah I see. How does all that quoting within the capture group work?

atteggiani: Yeah, the only downside is that.
You would need 3 separate "quoted" parts because you need double quotes for `$groups_regex` (you need to expand its value).
We might be able to get away with using only double quotes though:
```
groups=$(yq "${{ steps.yq.outputs.filter }} | capture(\"$groups_regex\")")
```

I usually try not to, especially when you are also substituting a `${{}}` github vaiable, to avoid some weird errors in case there are any characters within the variable that bash would expand (for example if `${{ steps.yq.outputs.filter }}` contained any `$` or `!` characters).

In this case, however, we should be safe anyway.

Using three different quoted parts would never give problems (but I understand is a bit harder to read).

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/235/commits/8e74ce86e9a0b0ffbcf1fbb0e2ebf97373370f42, thanks

atteggiani (approved):
Looks good!

PR #243, opened 2025-03-07
Undeploy: Export DEPLOYMENT_TARGET before removing environments

### LINKED ISSUE #242: Deployment Removal Workflow silently failing due to 'DEPLOYMENT_TARGET' being unset
See https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/13710579242/job/38346041442

Since `DEPLOYMENT_TARGET` is used within `spack.definitions[].when` clauses, `spack` now expects (and checks) for this environment variable in Prerelease when running through environments to remove. 

Export this variable, as is done in `deploy-2-start.yml`.

CodeGat:
Closes #242

## Background

A curious regression introduced in #218, is that now that we can (and have, in `ACCESS-TEST` Prereleases) use the set-at-install-time environment variable `DEPLOYMENT_TARGET` in our `spack.yaml`s, `spack` expects it to be set. Interestingly, `spack` now errors out when it can't find it, since it loads config from all existing environments `spack.yaml` while uninstalling some environments, including ones that require `DEPLOYMENT_TARGET`. 

## The PR

In this PR:
* export `DEPLOYMENT_TARGET` before undeploying environments 

## Testing

Tested uninstallation of a dummy enviroment in Prerelease without exporting `DEPLOYMENT_TARGET` which fails as expected.
Exported `DEPLOYMENT_TARGET` and then attempted uninstallation, which succeeded.

### DIFF
--- a/.github/workflows/undeploy-1-start.yml
+++ b/.github/workflows/undeploy-1-start.yml
@@ -68,6 +68,11 @@ jobs:
         id: undeploy
         run: |
           ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+
+          # Export vars.DEPLOYMENT_TARGET
+          export DEPLOYMENT_TARGET="${{ vars.DEPLOYMENT_TARGET }}"
+          echo "DEPLOYMENT_TARGET exported as $DEPLOYMENT_TARGET"
+
           . ${{ steps.path.outputs.spack-config }}/spack-enable.bash
           envs=$(find ${{ steps.path.outputs.spack }}/../environments -type d -name '${{ inputs.version-pattern }}' -printf '%f ')
 


harshula (approved):
:crossed_fingers:

atteggiani (approved):
Looks good! 

Just a general comment on `DEPLOYMENT_TARGET` and the `spack.yaml` file. 
I am not sure of all the possible values of `DEPLOYMENT_TARGET` but I believe it would be something like `Gadi`, `Setonix`, etc.
Isn't there a way to have `spack` automatically detect the `DEPLOYMENT_TARGET` (from the hostname of the machine where it is run for example), if no explicit `DEPLOYMENT_TARGET` is set within the environment?

CodeGat:
That is true @atteggiani, and indeed users can specify that in their `spack.yaml` via `when: re.match("gadi.nci.org.au", hostname)`, but there may be a time where we have multiple spack instances outside of the `Gadi [Pre]Release` ones on Gadi. For example, maybe a `Test` spack instance or something, that has different requirements than Prerelease.

CodeGat:
See succeeding run: https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/13710579242/job/38348497257

PR #244, opened 2025-03-07
Draft PRs disable schema and package projection checks

### LINKED ISSUE #237: Remove schema checks on Draft
## Background

The restrictive schema has been in place for Prereleases as the original rationale was that Prereleases were a natural testing ground for eventual Releases. 

This has evolved slightly, with developers using Prereleases as sharable testing grounds with no intention of merging them, making the schema unnessecarily restrictive. 

There have been other pull requests to `build-cd` that have weakened the restrictions on the schema (such as #235 and access-nri/schema#45) or removed restrictions on Draft PRs (such as #226). 

As a logical next step, we are looking to remove all schema restrictions on Draft PRs, giving developers the flexibility to test what they want, if there is no intention to make it a Release. 

## Implementation Details

* Make `check-spack-yaml` job conditional on the draft status of the PR. 
  * Decide if we still need to do the modulefile dependency checks for Draft PRs. 
* Determine how comments will look for draft PRs - there is a possibilty that there is no longer a `spack.specs[0]`/`spack.definitions.ROOT_PACKAGE[0]`. 

Thanks @harshula for the suggestion. Pinging @anton-seaice and @dougiesquire so they know when this is done.

CodeGat:
Closes #237

## Background

Prereleases have evolved - sometimes they are used not to create a Release, but to have a sharable location to test out various builds. This means that the schema (and some of the checks) are unnessecarily restrictive. This PR removes schema checks and package projection checks for Draft PRs in Model Deployment Repositories. 

## Open Questions

Wouldn't mind @aidanheerdegen and @harshula s opinions on these questions:

* Say we have an environment that installs `mom5@development` with spack package hash `abcdefg`. If we create another environment that reuses this package and therefore it's hash, will we have a modulefile clash even with a modulefile name `{name}/development-{hash:7` (resolves to `mom5/development-abcdefg`)?
  * And is this actually a problem? The packages are identical - it's not like the earlier case where we had `{name}/development` clashing the other, non-identical builds of `mom5@development`. 

## Testing

Using `ACCESS-TEST` with branch `237-skip-check-spack-yaml-check-on-draft_TEST` (which is the PR branch with all access-nri workflow refs replaced with the above branch). Did this with `sed -E 's|uses: access-nri/build-cd/(.+)@v4|uses: access-nri/build-cd/$1@237-skip-check-spack-yaml-check-on-draft_TEST|g' -i .github/workflows/*`

See https://github.com/ACCESS-NRI/ACCESS-TEST/pull/31 - this contains a run of tests that validate:

* Commit-invoked Draft PRs skip schema and package projection checks :heavy_check_mark: 
* `!redeploy`-invoked Draft PRs skip schema and package projection checks :heavy_check_mark: 
* Commit-invoked Non Draft PRs do NOT skip schema and package projection checks (and `!bump` continues to work) :heavy_check_mark: 
* `!redeploy`-invoked Non Draft PRs do NOT skip schema and package projection checks :heavy_check_mark: 

This verifies that this PR is working as expected.

### DIFF
--- a/.github/workflows/deploy-1-setup.yml
+++ b/.github/workflows/deploy-1-setup.yml
@@ -173,6 +173,9 @@ jobs:
           ref: ${{ inputs.deployment-ref }}
 
       - name: Validate ACCESS-NRI spack.yaml Restrictions
+        if: >-
+          (github.event_name == 'pull_request' && !github.event.pull_request.draft) ||
+          (github.event_name == 'issue_comment' && !github.event.issue.draft)
         uses: access-nri/schema/.github/actions/validate-with-schema@main
         with:
           schema-version: ${{ vars.SPACK_YAML_SCHEMA_VERSION }}
@@ -185,7 +188,7 @@ jobs:
 
       - name: Check '@latest' version usage
         if: >-
-          (github.event_name == 'pull_request' && !github.event.pull_request.draft) || 
+          (github.event_name == 'pull_request' && !github.event.pull_request.draft) ||
           (github.event_name == 'issue_comment' && !github.event.issue.draft)
         run: |
           if [[ "${{ steps.current.outputs.root-spec-ref }}" == "latest" ]]; then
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
> * And is this actually a problem?

What happens when a PR is closed or merged? Will the module file be deleted by one environment while it is still being used by another?

CodeGat:
No, I've found out that modulefiles aren't handled by `spack` outside of creating them...gotta remove them ourselves!

CodeGat:
I've verified that this PR works as it stands currently, so I'm setting it to Ready for Review - test details are in the PR description.

aidanheerdegen:
What happens with `!bump` in a draft PR?

CodeGat:
It'd still bump the version. It doesn't care what state the PR is in

aidanheerdegen:
> It'd still bump the version. It doesn't care what state the PR is in

Does that make sense for a draft PR which isn't checking the version stuff?

aidanheerdegen:
> will we have a modulefile clash even with a modulefile name? And is this actually a problem? 

Yes, and given we're going to have to create logic to remove the module files it is a problem if we remove a modulefile that is being used by another PR build.

So we should probably prepend the PR namespace to the module file projection automagically so we can easily and safely clean them up when the pre-release is cleaned up.

.github/workflows/deploy-1-setup.yml
```diff
@@ -236,40 +239,47 @@ jobs:
...
-              DEP_VER=${{ steps.current.outputs.root-spec-ref }}
-            else
-              # Capture the section after '@git.' or '@' (if it's not a git-attributed version) and before a possible '=' for a given dependency.
-              # Ex. '@git.2024.02.11' -> '2024.02.11', '@access-esm1.5' -> 'access-esm1.5', '@git.2024.05.21=access-esm1.5' -> '2024.05.21'
-              DEP_VER=$(yq ".spack.packages.\"$DEP\".require[0] | match(\"^@(?:git.)?([^=]*)\").captures[0].string" spack.yaml)
-            fi
+            # Capture the section after '@git.' or '@' (if it's not a git-attributed version) and before a possible '=' for a given dependency.
+            # Ex. '@git.2024.02.11' -> '2024.02.11', '@access-esm1.5' -> 'access-esm1.5', '@git.2024.05.21=access-esm1.5' -> '2024.05.21'
+            package_ver=$(yq ".spack.packages.\"$DEP\".require[0] | match(\"^@(?:git.)?([^=]*)\").captures[0].string" spack.yaml)
+            expected_projection_prefix="{name}/$package_ver"
```
aidanheerdegen, line 274: The internet is wonderful!

https://regex101.com/r/ezLyAr/1

.github/workflows/deploy-1-setup.yml
```diff
@@ -236,40 +239,47 @@ jobs:
...
 
-            # Get the version from the module projection, for comparison with DEP_VER
-            # Projections are of the form '{name}/VERSION[-{hash:7}]', in which we only care about VERSION. For example, '{name}/2024.11.11', or '{name}/2024.11.11-{hash:7}'
-            MODULE_NAME=$(yq ".spack.modules.default.tcl.projections.\"$DEP\"" spack.yaml)
-            MODULE_VER="${MODULE_NAME#*/}"  # Strip '{name}/' from '{name}/VERSION' module, even if VERSION contains '/'
-            MODULE_VER="${MODULE_VER%%-\{hash:7\}}"  # Strip a potential '-{hash:7}' appendix from the VERSION, since we won't have that in the DEP_VER
+            # Projections must start with '{name}/VERSION'. For example, '{name}/2024.11.11', or '{name}/2024.11.11-{hash:7}'
```
aidanheerdegen, line 276: It isn't clear to me that '{name}/2024.11.11-{hash:7}' is supported given the `expected_projection` doesn't include the `-{hash:7}` suffix.

aidanheerdegen (approved):
LGTM

.github/workflows/deploy-1-setup.yml
CodeGat: Indeed, regex101 is a mainstay for me too :laughing:

.github/workflows/deploy-1-setup.yml
CodeGat: It's a little further down, in the conditional:
```bash
if [[ "$current_projection" == "$expected_projection_prefix"* ]]; then
```
Note the `*` glob after the expected projection prefix, essentially being a startsWith check. 

I'd done it with globs instead of regex because it was a little bit simpler to set up, but  happy to change it to something a bit more understandable!

CodeGat:
> > It'd still bump the version. It doesn't care what state the PR is in
> 
> Does that make sense for a draft PR which isn't checking the version stuff?

I mean, not really. But it'd have to be a conscious choice to `!bump` the version in a Draft, which I think we should keep as functionality. Maybe someone wants to draft the PR as it's not ready to merge (yet) and wanted to bump the version anyway. 

> So we should probably prepend the PR namespace to the module file projection automagically so we can easily and safely clean them up when the pre-release is cleaned up.

I think this should be a separate issue, because it might have implications for users outside of this PR.

.github/workflows/deploy-1-setup.yml
aidanheerdegen: All good! Thanks for pointing out what I'd missed.

Issue #155, opened 2024-10-30
`ci-closed.yml` Doesn't Clean Up Modulefiles


CodeGat:
If you run a `module avail` after `module use /g/data/vk83/prerelease/modules`, you'll see that there are modulefiles for modules that don't exist anymore. Augment the `undeploy-2-start.yml` workflow so that the modulefiles are removed as well as the `spack env`s.

CodeGat:
I've copied over the `0.22 Prerelease` for some testing. 

* `spack module tcl rm` only removed a subset of the existing modulefiles - much less than expected for how many modulefiles there were (almost 20s of scrolling in `module avail`!). So I don't think this is the solution. 
* `spack module tcl refresh` did nothing. 
* `spack module tcl refresh --delete-tree` led to only a very small subset of the modulefiles remaining - much less than expected for the amount of environments there were. 

I think we should leave `spack module tcl` alone...

CodeGat:
After a chat with the spack developers (with @harshula and @aidanheerdegen), we noted that we could put the modules within the environment itself, similar to what we do the spack package repos. 

Something like:

```yaml
modules:
  default:
    roots:
      tcl: $env/modules
```

in `spack-config` would mean that modules would be created in the environment, and deleted when we `spack env rm` the environment. One thing that we would need to update is the module magic @aidanheerdegen created, to instead look in `/g/data/vk83/prerelease/apps/spack/*/environments/*/modules`. This sounds like what we want, and we won't have a 5000-line module index where 80% of the modules no longer exist 😆 We would have to double check that user instances expect such a thing.

CodeGat:
I shall test this in OM2 at the environment level, even though we are looking for this change to be targeted at `spack-config`

CodeGat:
I've tested it in https://github.com/ACCESS-NRI/ACCESS-OM2/pull/163 @aidanheerdegen and @harshula - it looks good! Modules were accessible by `module use`ing the environments modules directory, and were removed when the PR was cleaned up. I've got more details in the linked PR, especially regarding the arch folder. If this looks good, the next step is to open an issue in `spack-config`, link it to this issue, and add the config there.

The only bit I'm not sure of is if the module root is `$env/modules`, what happens if things are installed outside of an environment? `build-cd` is fine, but more regarding the shared spack instance.

CodeGat:
[...]

PR #248, opened 2025-03-14
Only Do Some Prerelease Modifications if the Root Spec is as expected

### LINKED ISSUE #247: Prerelease modifications leading to forced installation of original root SBR
See spack.yaml `spack.specs[0]` that caused the error https://github.com/ACCESS-NRI/ACCESS-OM3/pull/64/files#diff-e8582e74fa156f4e5729a850e52b24f2fde2d815c2c9c360f88c4cf90db851abR14 as well as the the error here https://github.com/ACCESS-NRI/ACCESS-OM3/actions/runs/13849168009/job/38753367324#step:10:268 and the actual cause of failure here: https://github.com/ACCESS-NRI/ACCESS-OM3/actions/runs/13849168009/job/38753367324#step:7:2

The issue is that when we are modifying the prerelease spack.yaml so that we don't use the (future) deployment version specified in the `spack.specs[0]`, we replace `spack.specs[0]` with the `inputs.root-sbd` provided by the caller. This is fine usually, but gives us a problem when Draft PRs are using a different `spack.specs[0]` for testing. 

Fix this so it uses `steps.yq.outputs.spack-root-spec-name` rather than `inputs.root-sbd`.

CodeGat:
Closes #247 

## Background

We do modifications to the Prerelease spack.yaml so that the `@git.VERSION` is removed, since it is a tag that doesn't yet exist; and update the modulefile so it is in the form `{name}/prX-Y`.

However, if users change the root spec to be something else (say, for testings sake), the prerelease modification will replace `spack.specs[0]` with the original `inputs.root-sbd`, removing their changes.

This PR does the `spack.specs[0]` modification only if the intended root-sbd (`inputs.root-sbd`)  is the same as the actual one (`steps.yq.outputs.root-spec-name`).

## Open Questions

* Is changing the root SBD something that we should support, anyway? Maybe only in drafted PRs? It kind of goes against the 'primacy of the root spec' thing that we have going.
  * As an aside, a user can always just have a secondary spec to install stuff that isn't the root spec, although it will increase the install time. 
* Cleaning up modulefiles after this may be difficult, although that is [an open issue](https://github.com/ACCESS-NRI/build-cd/issues/155) currently. 

## Testing

* Manually ran through varying values for `steps.yq.outputs.root-spec-name` on a local `spack.yaml` and verified the correctness of the updates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,7 +200,7 @@ jobs:
       deployment-version: ${{ needs.defaults.outputs.prerelease-version }}
       prerelease-compare-ref: ${{ needs.defaults.outputs.base-ref }}
       spack-manifest-path: ./spack.yaml
-      spack-manifest-root-sbd: ${{ needs.defaults.outputs.root-sbd }}
+      expected-root-spec-name: ${{ needs.defaults.outputs.root-sbd }}
     secrets: inherit
 
   redeploy-post:
--- a/.github/workflows/deploy-1-setup.yml
+++ b/.github/workflows/deploy-1-setup.yml
@@ -40,12 +40,12 @@ on:
         description: |
           Relative path in the Model Deployment Repository that contains the spack manifest file.
           Usually a spack.yaml file.
-      spack-manifest-root-sbd:
+      expected-root-spec-name:
         type: string
         required: true
         description: |
-          Within the spack manifest, the overarching spack bundle that contains all other packages.
-          Usually the first entry in the .spack.specs section of the manifest.
+          The expected overarching spack bundle that contains all other packages, set in the caller repository.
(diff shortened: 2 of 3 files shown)


aidanheerdegen:
> * Is changing the root SBD something that we should support, anyway? Maybe only in drafted PRs?

Definitely only for drafted PRs. If they need to change the root spec they should change `inputs.root-sbd`. Otherwise it's chaos.

aidanheerdegen (dismissed):
I approve, but I think we should only allow this for Draft.

I'm concerned we end up having a PR that is clearly un-deployable that might be passing tests.

So if we don't already, we should test in the `Release` version that `spec[0]` matches the root-sbd, throw an error, and let them know that is only permissible in draft PRs.

CodeGat:
I'll add that in now, before merging

aidanheerdegen (approved):
LGTM, and thanks.

PR #252, opened 2025-03-18
Remove workflow-level `concurrency` due to job cancellations

### LINKED ISSUE #251: Concurrency Only Lets One Job Have `pending` Status, `cancelled` Others
## Background

See https://github.com/ACCESS-NRI/ACCESS-ESM1.6/actions/runs/13846203559

This 'Higher Priority Job' cancellation started happening around 20 days ago, when people either did a `!redeploy` or commit to a PR that uses the same environment, at the same time. The only change from around then was an update to the concurrency key. We wanted access to the environment to still be one at a time per repo, but not to cancel existing jobs.

I went through the GitHub docs and they'd updated the docs a couple months ago to clarify that concurrency groups can only have one pending job at a time. So:
```
job1: in-progress
job2: pending
+
job3
=
job1: in-progress
job2: CANCELLED
job3: pending
```

So it's basically, `cancel-in-progress: false` is the same as `cancel-in-progress: true` except that you get ONE job that is in pending instead of being cancelled outright, but if you start a third job then it gets cancelled anyway.

## Implementation Plan

* Attempt removing the `concurrency` aspect and see if `spack` can handle the multiple installations at once. It's basically doing that at the organisation level, anyway.

CodeGat:
Closes #251

## Background

The concurrency that GitHub offers only allows for a single pending job while another is in progress. If another job in the same concurrency job is created, the previously `pending` job would be `cancelled`, and the new job would be `pending`. This isn't good for repositories with high levels of activity.

## The PR:

In this PR:

* Remove the `concurrency` key and see if `spack` can handle multiple installations at the same time at the repository level. We already allow multiple installations at the organisation level.

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -1,5 +1,4 @@
 name: Deploy Start
-concurrency: ${{ inputs.deployment-target }} ${{ inputs.deployment-type }}
 on:
   workflow_call:
     inputs:


aidanheerdegen (approved):
Let's try this thing!

PR #253, opened 2025-03-18
Update remote temp spack manifest name to prevent ambiguity

CodeGat:
See https://github.com/ACCESS-NRI/ACCESS-OM3/pull/64, specifically https://github.com/ACCESS-NRI/ACCESS-OM3/pull/64#issuecomment-2731653522

## Background

When a large amount of prereleases from the same repository are created around the same time, there is a possibility of the remote, temporarily stored `MODEL.spack.yaml` to be overwritten before the `spack env create` call. 

## The PR

In this PR:

* We further disambiguate the temporary remote file name with the job id number - for example, `access-om3-82974392347.spack.yaml`, rather than just `access-om3.spack.yaml`


atteggiani (approved):
LGTM!

PR #256, opened 2025-03-20
`spack-packages` concurrent access issue

### LINKED ISSUE #255: Concurrent Access Issues
When multiple builds that all change `spack-packages` are invoked, there is a possibility that they are locked from changing the `spack-packages` version, which leads to the possibility of the build using the wrong version of `spack-packges`. 

In this case, we need to separate installations of `spack-packages` down to the environment level, and inject the `spack.yaml` with a `repos` section that uses the following path to `spack-packages`: `$SPACK_ROOT/spack-packages/MODEL-prX-Y/spack-packages`.

This change will be made in conjunction with `spack-config` changes that sets the `config.db_timeout: 600` rather than `60`.

CodeGat:
Closes #255

## Background 

There is a possibility that multiple deployments to an instance of `spack` may lead to `spack-packages` being updated (and locked) during install time. Even though spack actually does copy the `packages.py` relevant to to the install prior to install, there is still a possibility that the version is changed before the `spack env create` call. 

The solution is to have an instance of `spack-packages` per commit, that each uses. Since the repository is only 1MB, this shouldn't be too much space. We link the commit-specific `spack-packages` in the `spack.yaml` at install-time via a `repos:: ["COMMMIT_SPECIFIC_SPACK_PACKAGES", "BUILTINS"]` section.

We also remove the `spack-packages` repositories when we close the PR. 

## The PR

In this PR:

- **Deployment: Handle having either a single spack-packages repo (for Release) or multiple (for Prerelease)**
- **Deployment: Add prerelease-specific repos:: section to spack.yaml**
- **Undeployment: Add new step to undeploy workflow that removes all spack-packages repos given by glob inputs.version-pattern**

## Testing

This branch is tested on `ACCESS-TEST`, specifically https://github.com/ACCESS-NRI/ACCESS-TEST/pull/33. It uses the https://github.com/ACCESS-NRI/build-cd/tree/255-spack-packages-concurrent-access-issue_TEST branch, which replaces all `@vX` references with `255-spack-packages-concurrent-access-issue_TEST`, which means that the entire pipeline uses the same commit, rather than just the entrypoints. 

### Runtime Modifications Test

Verified that the runtime spack-packages was successfully injected, having the form:
```yaml
spack:
  # ...
  repos::
  - /g/data/vk83/testing/apps/spack/0.22/spack-packages/access-test-pr33-3/spack-packages
  - $spack/var/spack/repos/builtin
```

### `spack-packages` Creation Logic During PR With Branch Ref

Success, see https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14025657298/job/39263794979?pr=33#step:10:33

### `spack-packages` Creation Logic During PR With Tag Ref

Success, see https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14025872250/job/39264344516?pr=33#step:10:35

### `spack-packages` Cleanup Logic During PR

Success, see https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14026246926/job/39265322646#step:6:1

### `spack-packages` Creation Logic During Release

Success, the `spack-packages` used is the one under `.../0.22/spack-packages`, rather than a child directory, see https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14026652957/job/39266451218#step:10:35

### DIFF
--- a/.github/actions/get-deploy-paths/action.yml
+++ b/.github/actions/get-deploy-paths/action.yml
@@ -48,14 +48,14 @@ runs:
   using: composite
   steps:
     - name: Validate Inputs
-      id: valid
+      id: validate
       shell: bash
       env:
         ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS: Release Prerelease
       run: |
         # Check that the inputs.spack-installs-root-path exists - it's usually a var
         if [ -z '${{ inputs.spack-installs-root-path }}' ]; then
-          echo '::error::`spack-installs-root-path` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment. Check Environment vars.'
+          echo '::error::inputs.spack-installs-root-path does not exist in ${{ github.repository }}s ${{ inputs.deployment-environment }} environment. Check Environment vars.'
           exit 1
         fi
 
@@ -77,8 +77,8 @@ runs:
         echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
         echo "spack-packages-root=$root/spack-packages" >> $GITHUB_OUTPUT
 
-        if [[ '${{ steps.valid.outputs.type }}' == Release ]]; then
+        if [[ '${{ steps.validate.outputs.type }}' == Release ]]; then
           echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
-        elif [[ '${{ steps.valid.outputs.type }}' == Prerelease ]]; then
+        elif [[ '${{ steps.validate.outputs.type }}' == Prerelease ]]; then
           echo "spack-packages=$root/spack-packages/${{ inputs.spack-environment }}/spack-packages" >> $GITHUB_OUTPUT
         fi
--- a/.github/actions/get-deploy-paths/action.yml
+++ b/.github/actions/get-deploy-paths/action.yml
@@ -16,8 +16,13 @@ inputs:
   deployment-environment:
     type: string
     required: true
-    # Used for logging, essentially
     description: Name of the GitHub deployment target environment
+  spack-environment:
+    type: string
+    required: true
+    description: |
+      Spack environment name that is used for this deployment.
+      Used to construct Prerelease spack-packages path, which is demarcated by the environment name.
 outputs:
   root:
     description: |
@@ -30,23 +35,50 @@ outputs:
   spack-config:
     description: Path to the ACCESS-NRI/spack-config repository associated with the install of spack.
     value: ${{ steps.path.outputs.spack-config }}
+  spack-packages-root:
+    description: |
+      Path to the folder containing all ACCESS-NRI/spack-packages repositories used in the install of spack.
+      For Release inputs.deployment-environment, this is the repository itself.
+      For Prerelease inputs.deployment-environment, this is the folder containing multiple spack-packages repositories.
+    value: ${{ steps.path.outputs.spack-packages-root }}
   spack-packages:
-    description: Path to the ACCESS-NRI/spack-packages repository associated with the install of spack.
+    description: Path to the ACCESS-NRI/spack-packages repository associated with the environment created in the install of spack.
     value: ${{ steps.path.outputs.spack-packages }}
 runs:
   using: composite
   steps:
-    - name: Get ${{ inputs.deployment-environment }} Remote Paths
-      id: path
+    - name: Validate Inputs
+      id: valid
       shell: bash
+      env:
+        ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS: Release Prerelease
       run: |
+        # Check that the inputs.spack-installs-root-path exists - it's usually a var
         if [ -z "${{ inputs.spack-installs-root-path }}" ]; then
           echo '::error::`spack-installs-root-path` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment. Check Environment vars.'
           exit 1
         fi
 
+        # We want the type of deployment environment, which is the last word in the string.
+        type=$(echo "${{ inputs.deployment-environment }}" | rev | cut -d ' ' -f 1 | rev)
+        if [[ ! "${{ env.ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS}}" =~ "$type" ]]; then
+          echo "::error::inputs.deployment-environment must be one of ${{ env.ACCEPTABLE_ENVIRONMENT_TARGETS }}."
+          exit 1
+        fi
+        echo "type=$type" >> $GITHUB_OUTPUT
+
+    - name: Get ${{ inputs.deployment-environment }} Remote Paths
+      id: path
+      shell: bash
+      run: |
         root=${{ inputs.spack-installs-root-path }}/${{ inputs.spack-version }}
         echo "root=$root" >> $GITHUB_OUTPUT
         echo "spack=$root/spack" >> $GITHUB_OUTPUT
         echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
-        echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
+        echo "spack-packages-root=$root/spack-packages" >> $GITHUB_OUTPUT
+
+        if [[ "${{ steps.valid.outputs.type }}" == "Release" ]]; then
+          echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
+        elif [[ "${{ steps.valid.outputs.type }}" == "Prerelease" ]]; then
+          echo "spack-packages=$root/spack-packages/${{ inputs.spack-environment }}/spack-packages" >> $GITHUB_OUTPUT
+        fi
--- a/.github/actions/get-deploy-paths/action.yml
+++ b/.github/actions/get-deploy-paths/action.yml
@@ -54,15 +54,15 @@ runs:
         ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS: Release Prerelease
       run: |
         # Check that the inputs.spack-installs-root-path exists - it's usually a var
-        if [ -z "${{ inputs.spack-installs-root-path }}" ]; then
+        if [ -z '${{ inputs.spack-installs-root-path }}' ]; then
           echo '::error::`spack-installs-root-path` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment. Check Environment vars.'
           exit 1
         fi
 
         # We want the type of deployment environment, which is the last word in the string.
-        type=$(echo "${{ inputs.deployment-environment }}" | rev | cut -d ' ' -f 1 | rev)
-        if [[ ! "${{ env.ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS}}" =~ "$type" ]]; then
-          echo "::error::inputs.deployment-environment must be one of ${{ env.ACCEPTABLE_ENVIRONMENT_TARGETS }}."
+        type=$(echo '${{ inputs.deployment-environment }}' | rev | cut -d ' ' -f 1 | rev)
+        if [[ ! '${{ env.ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS}}' =~ "$type" ]]; then
+          echo '::error::inputs.deployment-environment must be one of ${{ env.ACCEPTABLE_ENVIRONMENT_TARGETS }}.'
           exit 1
         fi
         echo "type=$type" >> $GITHUB_OUTPUT
@@ -77,8 +77,8 @@ runs:
         echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
         echo "spack-packages-root=$root/spack-packages" >> $GITHUB_OUTPUT
 
-        if [[ "${{ steps.valid.outputs.type }}" == "Release" ]]; then
+        if [[ '${{ steps.valid.outputs.type }}' == Release ]]; then
           echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
-        elif [[ "${{ steps.valid.outputs.type }}" == "Prerelease" ]]; then
+        elif [[ '${{ steps.valid.outputs.type }}' == Prerelease ]]; then
           echo "spack-packages=$root/spack-packages/${{ inputs.spack-environment }}/spack-packages" >> $GITHUB_OUTPUT
         fi
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -161,12 +161,15 @@ jobs:
             git clone https://github.com/ACCESS-NRI/spack-packages.git "${{ steps.path.outputs.spack-packages }}"
           fi
 
-          git -C ${{ steps.path.outputs.spack-packages }} fetch --all --tags
+          cd ${{ steps.path.outputs.spack-packages }} || exit 1
+
+          git fetch --all --tags
 
           if git ls-remote --exit-code --heads origin ${{ steps.versions.outputs.packages }}; then  # If the ref is a branch
-            # Checkout the branch at the latest commit
+            echo "Checking out the branch at the latest commit"
             git checkout -B ${{ steps.versions.outputs.packages }} origin/${{ steps.versions.outputs.packages }}
           else  # checkout the tag/commit
+            echo "Checking out the tag/commit"
             git checkout ${{ steps.versions.outputs.packages }}
           fi
           EOT
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -149,16 +149,16 @@ jobs:
         run: |
           ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
           # Check that a suitable spack-packages location exists
-          if [ ! -d "${{ steps.path.outputs.spack-packages-root }}" ]; then
+          if [ ! -d '${{ steps.path.outputs.spack-packages-root }}' ]; then
             echo '::error::spack-packages does not exist in `${{ steps.path.outputs.root }}` for `${{ inputs.deployment-target }} ${{ inputs.deployment-type}}`'
             exit 1
           fi
 
           # In the case of a Prerelease, we need to create a new instance of spack-packages per commit, so there is no concurrent access problems.
-          if [ ! -d "${{ steps.path.outputs.spack-packages }}" ]; then
-            echo "Creating spack-packages instance for ${{ inputs.env-name }} under ${{ steps.path.outputs.spack-packages }}"
+          if [ ! -d '${{ steps.path.outputs.spack-packages }}' ]; then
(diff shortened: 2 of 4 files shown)


.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
-    description: Path to the ACCESS-NRI/spack-packages repository associated with the install of spack.
+    description: Path to the ACCESS-NRI/spack-packages repository associated with the environment created in the install of spack.
     value: ${{ steps.path.outputs.spack-packages }}
 runs:
   using: composite
   steps:
-    - name: Get ${{ inputs.deployment-environment }} Remote Paths
-      id: path
+    - name: Validate Inputs
+      id: valid
```
atteggiani: Very minor, but I would use `validate` or `validate-inputs` instead of `valid` as the step ID.

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
       shell: bash
+      env:
+        ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS: Release Prerelease
       run: |
+        # Check that the inputs.spack-installs-root-path exists - it's usually a var
         if [ -z "${{ inputs.spack-installs-root-path }}" ]; then
```
atteggiani:
```suggestion
        if [ -z '${{ inputs.spack-installs-root-path }}' ]; then
```
Minor, but for context directly coming from inputs, I would use single quotes to avoid any unexpected behaviour in case the input contains characters that bash would expand (like `!` or `$`).

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
           echo '::error::`spack-installs-root-path` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment. Check Environment vars.'
           exit 1
         fi
 
+        # We want the type of deployment environment, which is the last word in the string.
+        type=$(echo "${{ inputs.deployment-environment }}" | rev | cut -d ' ' -f 1 | rev)
```
atteggiani:
```suggestion
        type=$(echo '${{ inputs.deployment-environment }}' | rev | cut -d ' ' -f 1 | rev)
```
Same as above.

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
+        if [[ ! "${{ env.ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS}}" =~ "$type" ]]; then
```
atteggiani:
```suggestion
        if [[ ! '${{ env.ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS}}' =~ "$type" ]]; then
```
This is not an input so it's more controllable, but in general I would use double quotes only if you want to expand whatever gets substituted in the github context. In this case you don't, so I would still use single quotes.

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
+          echo "::error::inputs.deployment-environment must be one of ${{ env.ACCEPTABLE_ENVIRONMENT_TARGETS }}."
```
atteggiani:
```suggestion
          echo '::error::inputs.deployment-environment must be one of ${{ env.ACCEPTABLE_ENVIRONMENT_TARGETS }}.'
```
Same as above

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
+      shell: bash
+      run: |
         root=${{ inputs.spack-installs-root-path }}/${{ inputs.spack-version }}
         echo "root=$root" >> $GITHUB_OUTPUT
         echo "spack=$root/spack" >> $GITHUB_OUTPUT
         echo "spack-config=$root/spack-config" >> $GITHUB_OUTPUT
-        echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
+        echo "spack-packages-root=$root/spack-packages" >> $GITHUB_OUTPUT
+
+        if [[ "${{ steps.valid.outputs.type }}" == "Release" ]]; then
```
atteggiani:
```suggestion
        if [[ '${{ steps.valid.outputs.type }}' == Release ]]; then
```
Same as above.
Also note that `"Release"` is the same as `Release`.

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
+          echo "spack-packages=$root/spack-packages" >> $GITHUB_OUTPUT
+        elif [[ "${{ steps.valid.outputs.type }}" == "Prerelease" ]]; then
```
atteggiani:
```suggestion
        elif [[ '${{ steps.valid.outputs.type }}' == Prerelease ]]; then
```
Same as above

.github/actions/get-deploy-paths/action.yml
```diff
@@ -30,23 +35,50 @@ outputs:
...
-      id: path
+    - name: Validate Inputs
+      id: valid
       shell: bash
+      env:
+        ACCEPTABLE_DEPLOYMENT_ENVIRONMENT_TARGETS: Release Prerelease
       run: |
+        # Check that the inputs.spack-installs-root-path exists - it's usually a var
         if [ -z "${{ inputs.spack-installs-root-path }}" ]; then
           echo '::error::`spack-installs-root-path` does not exist in `${{ github.repository }}`s `${{ inputs.deployment-environment }}` environment. Check Environment vars.'
```
atteggiani: Is the use of backticks `` ` `` deliberate? 
In `bash` backticks are used for command substitution. In this specific case they don't get expanded because the whole parameter is wrapped in single quotes `'`, but in general they're used to signify a command.
In this case the parts wrapped in backticks are not commands but rather paths or specific names. 
In this specific case I would use double quotes rather than backticks.

aidanheerdegen:
This looks like it is using the original `spack-packages`:

https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14025657298/job/39263794979?pr=33#step:11:45

```
==> [2025-03-24-13:07:13.555210] Reading config from file /g/data/vk83/testing/apps/spack/0.22/spack/etc/spack/defaults/packages.yaml
```
but later it is picking up the cloned version

https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/14025657298/job/39263794979?pr=33#step:11:213
```
==> [2025-03-24-13:09:37.258346] Installing /g/data/vk83/testing/apps/spack/0.22/spack-packages/access-test-pr33-3/spack-packages/packages/access-test/package.py to /g/data/vk83/testing/apps/spack/0.22/environments/access-test-pr33-3/.spack-env/repos/access.nri/packages/access-test
==> [2025-03-24-13:09:37.266453] Installing /g/data/vk83/testing/apps/spack/0.22/spack-packages/access-test-pr33-3/spack-packages/packages/access-test-component/package.py to /g/data/vk83/testing/apps/spack/0.22/environments/access-test-pr33-3/.spack-env/repos/access.nri/packages/access-test-component
```

So ... 👍

.github/workflows/deploy-2-start.yml
```diff
@@ -110,6 +112,13 @@ jobs:
...
 
           yq -i '${{ env.SPACK_YAML_MODULEFILE_PROJECTION_YQ }} = "{name}/${{ inputs.version }}"' spack.yaml
+
+          # Add the specific prerelease spack-packages repo to the environment
+          # We can't use yq as we are using a non-standard spack-specific '::' construct to ignore higher config scopes
+          echo '  repos::' >> spack.yaml
+          echo '  - ${{ steps.path.outputs.spack-packages }}' >> spack.yaml
+          echo '  - $spack/var/spack/repos/builtin' >> spack.yaml
+
           echo '::notice::Prerelease accessible as module `${{ steps.manifest.outputs.root-spec-name }}/${{ inputs.version }}`'
```
atteggiani: Same comment on backticks as above.

.github/workflows/deploy-2-start.yml
```diff
@@ -136,6 +145,35 @@ jobs:
           done
           EOT
 
+      - name: Update spack-packages repository
+        run: |
+          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          # Check that a suitable spack-packages location exists
+          if [ ! -d "${{ steps.path.outputs.spack-packages-root }}" ]; then
+            echo '::error::spack-packages does not exist in `${{ steps.path.outputs.root }}` for `${{ inputs.deployment-target }} ${{ inputs.deployment-type}}`'
```
atteggiani: Same comment on backticks as above.

.github/workflows/deploy-2-start.yml
```diff
@@ -136,6 +145,35 @@ jobs:
           done
           EOT
 
+      - name: Update spack-packages repository
+        run: |
+          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          # Check that a suitable spack-packages location exists
+          if [ ! -d "${{ steps.path.outputs.spack-packages-root }}" ]; then
```
atteggiani:
```suggestion
          if [ ! -d '${{ steps.path.outputs.spack-packages-root }}' ]; then
```
Same as above

.github/workflows/deploy-2-start.yml
```diff
@@ -136,6 +145,35 @@ jobs:
...
+            echo '::error::spack-packages does not exist in `${{ steps.path.outputs.root }}` for `${{ inputs.deployment-target }} ${{ inputs.deployment-type}}`'
+            exit 1
+          fi
+
+          # In the case of a Prerelease, we need to create a new instance of spack-packages per commit, so there is no concurrent access problems.
+          if [ ! -d "${{ steps.path.outputs.spack-packages }}" ]; then
```
atteggiani:
```suggestion
          if [ ! -d '${{ steps.path.outputs.spack-packages }}' ]; then
```
Same as above

.github/workflows/deploy-2-start.yml
```diff
@@ -136,6 +145,35 @@ jobs:
...
+            echo "Creating spack-packages instance for ${{ inputs.env-name }} under ${{ steps.path.outputs.spack-packages }}"
```
atteggiani:
```suggestion
            echo 'Creating spack-packages instance for ${{ inputs.env-name }} under ${{ steps.path.outputs.spack-packages }}'
```
Same as above

.github/workflows/deploy-2-start.yml
```diff
@@ -136,6 +145,35 @@ jobs:
...
+            mkdir -p ${{ steps.path.outputs.spack-packages }}
+            git clone https://github.com/ACCESS-NRI/spack-packages.git "${{ steps.path.outputs.spack-packages }}"
```
atteggiani:
```suggestion
            git clone https://github.com/ACCESS-NRI/spack-packages.git '${{ steps.path.outputs.spack-packages }}'
```
Same as above.
And, in general, I would be consistent on the quoting: if you don't quote it in the line above, I wouldn't quote it here either. Or I would single-quote it in both lines.

.github/workflows/undeploy-1-start.yml
```diff
@@ -63,9 +64,18 @@ jobs:
             ${{ secrets.HOST }}
             ${{ secrets.HOST_DATA }}
 
-      - name: Undeploy
+      - name: Undeploy spack-packages
+        # ssh into deployment environment, remove all the unneeded spack-packages installations
+        run: |
+          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<EOT
```
atteggiani:
```suggestion
          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
```
I don't think you need any `bash` expansion within the heredoc script, so I would escape the whole script

.github/workflows/undeploy-1-start.yml
```diff
@@ -86,10 +96,5 @@ jobs:
...
       - name: Undeploy Status Notifier
-        if: always()
-        run: |
-          if [[ "${{ steps.undeploy.outcome }}" == "success" ]]; then
-            echo "::notice::Deployment ${{ inputs.version-pattern }} was successfully removed from ${{ inputs.target }} ${{ inputs.type }}, found at ${{ steps.path.outputs.spack }}"
-          else
-            echo "::error::Deployment ${{ inputs.version-pattern }} couldn't be removed from ${{ inputs.target }} ${{ inputs.type }}, found at ${{ steps.path.outputs.spack }}. Please check manually."
-          fi
+        if: failure()
+        run: echo "::error::Deployment ${{ inputs.version-pattern }} couldn't be removed from ${{ inputs.target }} ${{ inputs.type }}, found at ${{ steps.path.outputs.spack }}. Please check manually."
```
atteggiani: Again, minor and most likely not needed in this case, but as a general rule I would avoid using double-quotes if you don't want to specifically expand `bash` characters within the string.
I would use multiple single-quoted parts, with escaped single-quoted characters for a printed single quote `'`:
```
echo '::error::Deployment ${{ inputs.version-pattern }} couldn'\''t be removed from ${{ inputs.target }} ${{ inputs.type }}, found at ${{ steps.path.outputs.spack }}. Please check manually.'
```

.github/workflows/deploy-2-start.yml
```diff
@@ -110,6 +112,13 @@ jobs:
           fi
 
           yq -i '${{ env.SPACK_YAML_MODULEFILE_PROJECTION_YQ }} = "{name}/${{ inputs.version }}"' spack.yaml
+
+          # Add the specific prerelease spack-packages repo to the environment
+          # We can't use yq as we are using a non-standard spack-specific '::' construct to ignore higher config scopes
+          echo '  repos::' >> spack.yaml
+          echo '  - ${{ steps.path.outputs.spack-packages }}' >> spack.yaml
+          echo '  - $spack/var/spack/repos/builtin' >> spack.yaml
```
aidanheerdegen: Very minor point, but I think this is slightly more readable as a `HEREDOC` block with a single redirect to `spack.yaml`.

aidanheerdegen (commented):
LGTM (one *tiny* suggestion)

Thanks for all the testing info. Made it a lot easier to review.

.github/workflows/undeploy-1-start.yml
```diff
@@ -63,9 +64,18 @@ jobs:
...
 
-      - name: Undeploy
+      - name: Undeploy spack-packages
+        # ssh into deployment environment, remove all the unneeded spack-packages installations
+        run: |
+          ssh ${{ secrets.USER}}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<EOT
+          find "${{ steps.path.outputs.spack-packages-root }}" \
+            -mindepth 1 -maxdepth 1 \
+            -type d -name "${{ inputs.version-pattern }}" \
+            -exec rm -rfv {} +
```
atteggiani, line 74: Using `rm -rf` together with `${{ inputs. ....}}` is super scary, but I guess it cannot be fixed until we find a "safe-rm" alternative.

atteggiani (dismissed):
Looks good to me! 

I left a few comments but they're all minor, and mostly related to the same "pattern" (single-quotes vs double-quotes).
Feel free to ignore all of them if you don't think they're particularly relevant in this case.

.github/workflows/deploy-2-start.yml
CodeGat: Only issue is that heredocs can't be indented, right?

Edit: Seems fine, actually.

atteggiani: In case, you might be able to trim the initial spaces/tabs, with something like:
```bash
cat >> spack.yaml <<EOF | sed 's/^[ \t]*//'
...
...
EOF
```
But this would remove all leading spaces, so you would not be able to indent the `- ...` lines if you wanted.

atteggiani: Another trick to be able to indent, is to use a random character (for example `@` as your new separator between indentation to keep and what not to keep, so you could have something like:
```bash
cat >> spack.yaml <<EOF | sed 's/^[ \t]*@//'
@... #non-indented part
@... #non-indented part
@    ... # indented part
EOF
```
Not sure this makes it easier to read though 😆

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/a11205c4b6210b366837f40dbbdd767319bb204e - works fine, which was surprising - I guess github de-indents the script part when executing.

.github/actions/get-deploy-paths/action.yml
CodeGat: Yeah, I was intending to use the backticks as a kind a emphasis tool to note that what I was saying was part of the workflow inputs itself (like how I would say `inputs.spack-installs-root-path` on GitHub), but now remember that they are used for command substitution in bash! I always use `$()` :D

.github/actions/get-deploy-paths/action.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/ff322305c6f5030f65d750a3341e4a7da8d4c80a

.github/actions/get-deploy-paths/action.yml
CodeGat: Removed in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/ff322305c6f5030f65d750a3341e4a7da8d4c80a

.github/workflows/deploy-2-start.yml
CodeGat: Removed in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/296a607621b495b33b666ab9dc7ad95862d00446

.github/workflows/deploy-2-start.yml
CodeGat: Removed in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/296a607621b495b33b666ab9dc7ad95862d00446

.github/workflows/undeploy-1-start.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/256/commits/eaf61862bf9a4299431a5339e7ea8b8dd5b25317

.github/workflows/undeploy-1-start.yml
CodeGat: The good thing is that the inputs don't come from user input, but from a previous action.
And the use of `find` over a bare `rm -rf PATH` means that the action is scoped to `${{ steps.path.outputs.spack-packages-root }}` before the `-exec rm` is called.
And for `${{ steps.path.outputs.spack-packages-root`, I've validated that even in the case where we do strange things to the workflow input like have inputs of `" "`, we can't do any proper damage because the path is quoted in `find`, and therefore couldn't do something like `find / /spack-packages`

.github/workflows/undeploy-1-start.yml
atteggiani: You solution was even better! 😆

atteggiani (approved):
LGTM!

aidanheerdegen (approved):
LGTM2

PR #258, opened 2025-03-21
Run both projection checks despite failure, run package projections on push (CD)

### LINKED ISSUE #257: Run package projection checks even if root spec checks failed
See https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/65.

Essentially, the package projection checks didn't run as the spec projection check failed first. 

But there is an open question still - If the spec was fixed in a later commit, why didn't the packages check run in the later commit?

CodeGat:
Closes #257

## Background

The package projection checks aren't running. This is due to some checks only running on pull_request and issue_comment non-draft, but not push (which we signify with `inputs.deployment-type == 'Release'`)!

## In This PR

* Run both projection checks (where appropriate) even if an earlier one fails
* Run the package projection checks on push
* Simplify the boolean logic


atteggiani (approved):
LGTM!

PR #259, opened 2025-03-25
Fail fast when installing manifest

CodeGat:
## Background

When installing models, the cause of installation failure causes can be obscured by `spack` continuing to install other dependencies. With `--fail-fast`, spack will exit more quickly and the cause of failures will no longer require a lot of scrolling. 

## The PR

* Add `--fail-fast` to the `spack install` flags


harshula (approved):
:crossed_fingers:

PR #261, opened 2025-04-01
Add back variants after Prerelease Root Spec modifications

### LINKED ISSUE #260: Variants are stripped from root spec due to prerelease modifications
See https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/70#issuecomment-2767665510

The variants are being stripped from the root spec because the Prerelease spack.yaml modifications replace the spec with only the root spec name. So:

```yaml
spack:
  specs:
    - access-om2@git.2025.04.0 +debug
```
becomes:
```yaml
spack:
  specs:
    - access-om2
```
instead of:
```yaml
spack:
  specs:
    - access-om2 +debug
```

CodeGat:
Closes #260

## Background

When we are doing our prerelease modifications, we replace the root spec with only the root spec name (removing any `@git.VERSION` as required), but we don't add back any variants that may have been in the original root spec definition. More details are in the linked issue. 

In this PR, we do our replacements a bit more carefully - so if the original spec was `access-om2@git.2025.03.0 +debug`, the replaced spec would be `access-om2 +debug` rather than `access-om2`.

## The PR

- **Export optional variants from get-spack-root-spec action**
- **Instead of just replacing the entire root spec with the root spec name, include the variants as well**

## Testing

* Tested out the hellish regex on `regex101` - https://regex101.com/r/AujFTw/1
* Tested in https://github.com/ACCESS-NRI/ACCESS-TEST/pull/34, specifically:
  * Space-separated variants are preserved :heavy_check_mark: [here](https://github.com/ACCESS-NRI/ACCESS-TEST/pull/34#issuecomment-2768030477)
  * 'Squished' variants are preserved :heavy_check_mark: [here](https://github.com/ACCESS-NRI/ACCESS-TEST/pull/34#issuecomment-2768035289)
  * If there are no variants, they are not added :heavy_check_mark: [here](https://github.com/ACCESS-NRI/ACCESS-TEST/pull/34#issuecomment-2768053083)

### DIFF
--- a/.github/actions/get-spack-root-spec/action.yml
+++ b/.github/actions/get-spack-root-spec/action.yml
@@ -28,6 +28,11 @@ outputs:
       The spack version from the root spec in the spack manifest file.
       For example, 'release'.
     value: ${{ steps.spec.outputs.version }}
+  root-spec-variants:
+    description: |
+      The variants from the root spec in the spack manifest file.
+      For example: '~deterministic'.
+    value: ${{ steps.spec.outputs.variants }}
   yq-root-spec:
     description: |
       The yq filter for the root spec of the spack manifest file.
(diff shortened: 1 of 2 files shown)


.github/actions/get-spack-root-spec/action.yml
```diff
@@ -60,26 +65,29 @@ runs:
...
 
         # Example of captured groups from the above:
         # name: anything before `@`. Ex: access-om2
         # ref: anything after `@` or `@git.`; but before `=` (for =VERSION syntax) or ` `/`+`/`~` (for variants). Ex: 2025.01.0
         # version: anything after a `=`, but before ` `/`+`/`~` (for variants). Ex: release
-        
-        groups_regex='(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?.*'
+        # variants: anything at the end that contains ` `/`+`/`~` Ex. ~variant +debug some=value
+
+        groups_regex='(?<name>.+)@(?:git\\.)?(?<ref>[^=+~ ]+)(?:=(?<version>[^~+ ]+))?(?<variants>[~+ ].+)?'
```
aidanheerdegen, line 78: Thanks for the regex101 link. I tried some combinations, and it seemed fine, but man this makes me nervous.

Thinking back again to testing, would it be possible to have some CI on this repo that called some of the workflows contained here with test inputs and check answers are as expected? 

Doesn't have to happen before this is merged, but wanted to start that conversation.

.github/workflows/deploy-2-start.yml
```diff
@@ -106,9 +106,10 @@ jobs:
         # is being used to test something else.
         env:
           SPACK_YAML_MODULEFILE_PROJECTION_YQ: .spack.modules.default.tcl.projections.${{ steps.manifest.outputs.root-spec-name }}
+          OPTIONAL_VARIANTS: ${{ steps.manifest.outputs.root-spec-variants != 'null' && steps.manifest.outputs.root-spec-variants || '' }}
```
aidanheerdegen, line 109: It's cool that the testing exposed this requirement.

aidanheerdegen (approved):
LGTM

.github/actions/get-spack-root-spec/action.yml
CodeGat: Yeah, regex is always a bit scary. I've made an issue https://github.com/ACCESS-NRI/build-cd/issues/262 for at least testing of actions when `.github/actions/*` is modified. Workflow testing will be more difficult, but we could probably at least test the non-environment-using bits.

PR #271, opened 2025-05-16
Allow access to `spack-cc-*` files

### LINKED ISSUE #270: Provide Pre-release users access to `spack-cc*.out` log files
Can build-cd expose the `spack-cc*.out` files to the user, in some manner? I suspect the user would find all the `spack-cc*.out` files corresponding to the SPRs in the Packages/Specs section of the `spack.yaml` to be useful when debugging.

We've already had multiple instances where this was needed by users.

CodeGat:
Closes #270

## Background

Currently, debug spack files (`spack-cc-*`) are stored in the `~` directory of the service user, inaccessible to others. 

Once a model is installed, we should move these log files to somewhere accessible to all - like `$SPACK_ROOT/logs/`.

# The PR

- **deploy-2-start.yml: Move spack debug files into $SPACK_ROOT/logs**
- **create-deployment-spack: Create $SPACK_ROOT/logs folder**

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -203,6 +203,9 @@ jobs:
           spack env activate ${{ inputs.env-name }}
           spack --debug install --fail-fast --fresh ${{ vars.SPACK_INSTALL_PARALLEL_JOBS }} || exit $?
           spack module tcl refresh -y
+
+          # Move spack debug logs to $SPACK_ROOT/logs
+          mv ~/spack-cc-* ${{ steps.path.outputs.root }}/logs
           EOT
 
       - name: Export deployment target locations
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -203,6 +203,9 @@ jobs:
           spack env activate ${{ inputs.env-name }}
           spack --debug install --fail-fast --fresh ${{ vars.SPACK_INSTALL_PARALLEL_JOBS }} || exit $?
           spack module tcl refresh -y
+
+          # Move spack debug logs to $SPACK_ROOT/logs
+          mv ~/spack-cc-* ${{ steps.path.outputs.root }}/logs
```
aidanheerdegen, line 208: If we have simultaneous builds does this have the possibility of moving the wrong build logs?

aidanheerdegen: I wonder if we can be more specific about what log files to move?

Is this the code where the path is specified?

https://github.com/spack/spack/blob/6e98f88c51e41019a6d5d66aaf7d2fcd4927504b/var/spack/repos/spack_repo/builtin/packages/compiler_wrapper/cc.sh#L195-L196

In which case it looks like `SPACK_DEBUG_LOG_ID` is defined here:

https://github.com/spack/spack/blob/6e98f88c51e41019a6d5d66aaf7d2fcd4927504b/lib/spack/spack/build_environment.py#L467

Can we access this in the CI environment?

CodeGat: I didn't really think of any of the logs as being wrong/part of any particular build - I just wanted to move all logs to somewhere more accessible.

And nah, I'd just tested it and it seems like it's only set during the command (with `--debug` on) and isn't there afterwards.

aidanheerdegen: > I didn't really think of any of the logs as being wrong/part of any particular build - I just wanted to move all logs to somewhere more accessible.

Oh sorry, yeah my mistake. I didn't realise they're all going into the same location. Seems fine.

aidanheerdegen (approved):
LGTM

PR #273, opened 2025-05-16
Update `spack install` var to `vars.SPACK_INSTALL_ADDITIONAL_ARGS`

### LINKED ISSUE #272: Advanced Option: Allow specification of other `spack install` flags via `vars.SPACK_INSTALL_ADDITIONAL_ARGS`
## Background

As noted by @manodeep, it would be good to be able to specify `--source` builds for `spack install`, so he can interrogate the source code. 

We don't want to turn this on globally as the inode usage could be high. But we already have set a precedent of repo-level customisation of the `spack install` command, as `ACCESS-OM3` uses the `vars.SPACK_INSTALL_PARALLEL JOBS` (set to `--jobs 4`) to affect the amount of jobs the `spack install` command uses - the variable is essentially a flag! 

This should be generalised to `vars.SPACK_INSTALL_ADDITIONAL_ARGS`, an advanced, optional variable used as part of the `spack install` command that allows additional flags where required - with the caveat that this should be used sparingly, and could break that repositories `build-cd` if they aren't careful.

CodeGat:
Closes #272

## Background

Advanced users may want to set `spack install` flags on a repo-by-repo basis. We unknowingly offer this via the `vars.SPACK_INSTALL_PARALLEL_JOBS` variable, but it isn't a great name. 

## The PR

In this PR:

* Make the variable that holds additional flags in `spack install` a more generic name - rather than `vars.SPACK_INSTALL_PARALLEL_JOBS`, it should be `vars.SPACK_INSTALL_ADDITIONAL_ARGS`.

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -201,7 +201,7 @@ jobs:
           # Create environment and build model
           spack env create ${{ inputs.env-name }} ${{ vars.SPACK_YAML_LOCATION }}/${{ inputs.expected-root-spec-name }}-${{ github.run_id }}.spack.yaml
           spack env activate ${{ inputs.env-name }}
-          spack --debug install --fail-fast --fresh ${{ vars.SPACK_INSTALL_PARALLEL_JOBS }} || exit $?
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM

PR #274, opened 2025-05-21
Conditionally mv spack-cc-* files

CodeGat:
## Background

`spack-cc-*` files are not always created, so the current implementation of moving the log files fails. 

## The PR

* Attempt to `find` the `spack-cc-*` files first before `mv`ing them. 

## Testing

* Tested the updated command on the service user


dougiesquire (approved):
LGTM

PR #281, opened 2025-06-03
Clashing Artifact Fix

### LINKED ISSUE #279: Multiple Tools Deployed in One Run Lead to Clashing Artifact Names
## Background

See https://github.com/ACCESS-NRI/model-tools/actions/runs/15404847732/job/43348624580.

We use `deploy-[metadata|outputs].TARGET` as an artifact name. This has worked fine for model deployment repositories that only deploy one spack manifest in any given run. But for non-model deployment repositories that can deploy multiple manifests (or tools) in a single workflow run, the artifact names clash. 

## Implementation Plan

* See if we can get the `inputs.model` into the artifact name, of the form `deploy-MODEL-[metadata|outputs].TARGET`. For example, `deploy-gh-metadata.Gadi` or `deploy-access-om2-outputs.Setonix`.

CodeGat:
Closes #279

## Background

The artifact names clash when there are multiple deployments in a single run - like in the case where we deploy multiple tools in a single run. They share the `deploy-[outputs|metadata].TARGET` filename. 

In this PR, we update the artifacts to be demarcated also on `root-spec`, so they look like `deploy-SOFTWARE-[outputs|metadata].TARGET`, for example `deploy-access-om3-outputs.Gadi`. 

## The PR

- Update artifact names to `deploy-SOFTWARE-[metadata|outputs].TARGET`
- Update build database filepaths for deploy-*-[outputs|metadata].* files
- Update test infra for build database regarding new filepaths

## Testing

Ran through the updated pytest suite for the model database, and verified that the artifact paths were not referred to anywhere else outside of their `env` definitions.


aidanheerdegen (approved):
LGTM, but would be good to get @tmcadam to review the provenance touching aspects.

tmcadam (approved):
LGTM, I don't think this requires any changes on the server side of release provenance

PR #284, opened 2025-06-12
Simplify and Fix Cleanup Workflow

### LINKED ISSUE #283: Cleaning Up Prereleases Doesn't Clean Properly
While we were fixing up an inode issue, it was noted that our current clean up workflow is insufficent. When we checked with `spack gc --except-any-environment`, we noted that there was almost 400+ packages that were not part of an environment, but were still installed. Since we only work in environments, there must be an issue with the current workflow. 

## Implementation Plan

Since using `spack gc --except-any-environment` seemed to clean up packages that were unused (rather than uninstalling packages in an environment about to be deleted) we should just use this in our undeploy workflow. 

We will need to verify that packages that are currently being installed are still considered 'in an environment'.

CodeGat:
## Background

Seems that our existing cleanup workflow is insufficent. When we ran `spack gc --except-any-environment` in our Prerelease area, there were 400+ packages installed that didn't need to be. 

## In this PR

* Instead of uninstalling packages within an environment (which may have package dependencies on other, to-be-uninstalled environments), just remove the environment and then uninstall packages not part of any environment with the above command.

### DIFF
--- a/.github/workflows/undeploy-1-start.yml
+++ b/.github/workflows/undeploy-1-start.yml
@@ -77,12 +77,9 @@ jobs:
           envs=$(find ${{ steps.path.outputs.spack }}/../environments -type d -name '${{ inputs.version-pattern }}' -printf '%f ')
 
           for env in $envs; do
-            spack env activate $env
-            spack uninstall --dependents --yes-to-all --all
-            spack env deactivate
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
Seems a good approach to me

PR #285, opened 2025-06-19
Update versioning of `spack` and `spack-config` in `settings.json`

justinh2002:
Update versioning of `spack` and `spack-config` in `settings.json`, update to include `ACCESS-ISSM` in `README.md`


CodeGat:
Thanks for that @justinh2002 :)

If you don't mind, would you be able to update the version of the Release spack/spack-config as well?

CodeGat:
Thanks - the version of spack as well, if you don't mind - just so deployments to the release area aren't using a version of spack that doesn't support the deployment

CodeGat (approved):
Looks good!

PR #288, opened 2025-06-24
Update modulerc

aidanheerdegen:
Record updated modulerc which supports general software deployments to the `model-tools` subdirectory.

Uses the `model-tools` namespace to make discovery more straightforward.


CodeGat (approved):
Looks good, thank you for that!

PR #290, opened 2025-06-30
Move logs regardless of failure or success of deployment

### LINKED ISSUE #289: Debug logs only moved on success rather than all the time
We only move the debug logs into the user-accessible space if the build succeeds, which defeats the purpose of post-failure forensics. 

Update the workflow to always attempt to move the debug logs.

CodeGat:
Closes #289

## Background

Before, we didn't move the logs if there was a failure as the `mv` command was skipped. We now have this movement of logs as a separate, always executing step. 

## The PR

* Move copy of logs to it's own step that runs regardless of deployment failure/success

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -187,6 +187,7 @@ jobs:
           EOT
 
       - name: Deploy to ${{ inputs.deployment-target }} ${{ inputs.deployment-type }}
+        id: deploy
         # ssh into deployment environment, create and activate the env, install the spack manifest.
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
Happy to approve, but can we make an issue to look into [`logrotate`](https://www.digitalocean.com/community/tutorials/how-to-manage-logfiles-with-logrotate-on-ubuntu-22-04) (or similar) to manage the proliferation of logs.

CodeGat:
Thanks @aidanheerdegen, the issue is in #291

PR #292, opened 2025-06-30
Tar Prerelease Spack Environment Metadata Post-Install

### LINKED ISSUE #277: Inode use in prerelease builds
### Introduction

The pre-release functionality creates and retains all the builds from every `push` to a PR.

This is very useful for developers as it allows for easy comparison between the versions of the build in the PR. A downside is that it can consume a large number of HPC system resources, typically inodes, that are subject to quotas.

All the builds for a PR are deleted when the PR is closed or merged, but it isn't always convenient to close a PR when it is still being used for development.

Not all PRs need to keep all the builds, and in those cases it would be fine to delete those unwanted builds while the PR was still open, but there is no way to do so easily.

### Proposal

Add a ChatOps command, e.g. `!clean` that could be used to delete intermediate builds that are not required.

If the command could take options we could pass it the number of builds to retain (including the latest build), 
e.g.

- `!clean 1`: remove all builds but the most recent
- `!clean 5`: remove all builds but the most recent and four preceding builds
- `!clean`: no argument default to remove all builds but the most recent

It would be useful to report what was deleted and the number of inodes saved if that was simple to do.

CodeGat:
Closes #277

## Background

Spack environments have large inode requirements. It [has been found](https://github.com/ACCESS-NRI/build-cd/issues/277#issuecomment-3007165768) that tarring up the `.spack-env` directory inside the spack environment saves a large amount of inodes, and we can still perform environment-aware commands (most notably, `spack env rm` and `spack gc -E`).

For all future prereleases, we should tar up this directory post-install. Users who use the spack instance for development (very few at this point) can either `tar -xf .spack-env .spack-env.tar` before use, otherwise the majority of the infromation will be recreated on install.

## The PR 

- **Add step tarring spack environment metadata to save inodes**
- **Update get-deploy-paths to output spack environment location**
- **Replace all refs to env.SPACK_ENV_PATH to the new steps.path.outputs.spack-environment**
- **Update guidance on spack environment comment**

### DIFF
--- a/.github/actions/get-deploy-paths/action.yml
+++ b/.github/actions/get-deploy-paths/action.yml
@@ -3,22 +3,18 @@ description: Action that returns important paths for a spack install on a deploy
 author: Tommy Gatti
 inputs:
   spack-installs-root-path:
-    type: string
     required: true
     # For example, `some/apps/spack` is the root, `some/apps/spack/0.21/` is a specific spack deployment, `some/apps/spack/0.21/spack` is a specific spack install.
     description: Path to a directory within which all versions of spack are installed
   spack-version:
-    type: string
     required: true
     description: |
       Version of spack deployed.
       Used to construct a specific spack installation path, in conjunction with `spack-installs-root-path`.
   deployment-environment:
-    type: string
     required: true
     description: Name of the GitHub deployment target environment
   spack-environment:
-    type: string
     required: true
     description: |
       Spack environment name that is used for this deployment.
@@ -44,6 +40,9 @@ outputs:
   spack-packages:
     description: Path to the ACCESS-NRI/spack-packages repository associated with the environment created in the install of spack.
     value: ${{ steps.path.outputs.spack-packages }}
+  spack-environment:
+    description: Path to the spack environment folder for the given inputs.spack-environment.
+    value: ${{ steps.path.outputs.spack-environment }}
 runs:
   using: composite
   steps:
@@ -82,3 +81,5 @@ runs:
         elif [[ '${{ steps.validate.outputs.type }}' == Prerelease ]]; then
           echo "spack-packages=$root/spack-packages/${{ inputs.spack-environment }}/spack-packages" >> $GITHUB_OUTPUT
         fi
+
+        echo "spack-environment=$root/environments/${{ inputs.spack-environment }}" >> $GITHUB_OUTPUT
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -226,8 +226,6 @@ jobs:
       - name: Get metadata from ${{ inputs.deployment-target }} ${{ inputs.deployment-type}}
         id: metadata
         # Get the metadata from the spack environment regarding build database packages
-        env:
-          SPACK_ENV_PATH: ${{ steps.path.outputs.spack }}/../environments/${{ inputs.env-name }}
         run: |
           ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
           # Export vars.DEPLOYMENT_TARGET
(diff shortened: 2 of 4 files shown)


.github/actions/get-deploy-paths/action.yml
```diff
@@ -3,22 +3,18 @@ description: Action that returns important paths for a spack install on a deploy
 author: Tommy Gatti
 inputs:
   spack-installs-root-path:
-    type: string
```
aidanheerdegen, line 6: Just curious why the `string` type had been removed from a bunch of inputs?

.github/workflows/deploy-2-start.yml
```diff
@@ -270,22 +268,26 @@ jobs:
...
             ./root-spec-pkg-hash.${{ github.run_id }}
 
           echo "$(cat ./root-spec-pkg-hash.${{ github.run_id }})" >> $GITHUB_OUTPUT
           rm ./root-spec-pkg-hash.${{ github.run_id }}
 
+      - name: Tar spack environment metadata
+        if: inputs.deployment-type == 'Prerelease'
+        run: |
+          ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+          tar --remove-files --create --file ${{ steps.path.outputs.spack-environment }}.spack-env.tar ${{ steps.path.outputs.spack-environment }}/.spack-env
```
aidanheerdegen: Should this check the directory exists before trying to `tar` it up? We wouldn't want the workflow to fail if for some reason the directory wasn't created.

scripts/jinja_template/templates/deployment-comment-body.md.j2
```diff
@@ -23,6 +23,7 @@ module load {{ root_sbd }}/{{ prerelease_version}}
 When using the above modules, the binaries shall be on your `$PATH`.
 
 For advanced users, this Prerelease is also accessible on `{{ deployment.name }}` via `{{ deployment.spack_location }}` in the `{{ root_sbd }}-{{ prerelease_version }}` environment.
+Due to inode-saving measures, one will have to manually untar the environment metadata before environment activation with `tar -xf .spack-env .spack-env.tar`.
```
aidanheerdegen: Maybe mention special permissions are required to do this.

aidanheerdegen (changes requested):
A couple of small things, otherwise great.

.github/actions/get-deploy-paths/action.yml
CodeGat: Actions don't have input types like workflows - they are all `string`s. It was coming up as an error to have that key in there in my linter, so I removed them.

.github/workflows/deploy-2-start.yml
CodeGat: Good point.

scripts/jinja_template/templates/deployment-comment-body.md.j2
CodeGat: Will do!

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/292/commits/de70bf5619e5f9f5d19289f1d3c4c72fbca4e6c1

.github/workflows/deploy-2-start.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/292/commits/de70bf5619e5f9f5d19289f1d3c4c72fbca4e6c1

aidanheerdegen (dismissed):
LGTM!

aidanheerdegen:
So is this being tested live? I'd like to check out some examples of it working because I have trust issues.

CodeGat:
Thanks @aidanheerdegen - good shout. 

I tested this using ACCESS-TEST via the linked PR:

I tested the the tarring logic and [it worked](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16016144865/job/45183078228?pr=47) after a path separation fix, this is the structure we are left with:

```bash
$ ls -alh /g/data/vk83/prerelease/apps/spack/0.22/environments/access-test-pr47-2/
total 504K
drwxrws---+   2 tm70_ci vk83 4.0K Jul  2 14:37 .
drwxrws---+ 419 tm70_ci vk83  40K Jul  2 14:32 ..
-rw-rw----+   1 tm70_ci vk83  352 Jul  2 14:36 build-db-pkgs.json
-rw-rw----+   1 tm70_ci vk83 440K Jul  2 14:37 .spack-env.tar
-rw-rw----+   1 tm70_ci vk83  735 Jul  2 14:36 spack.location
-rw-rw----+   1 tm70_ci vk83 4.7K Jul  2 14:36 spack.lock
-rw-r--r--+   1 tm70_ci vk83 1.6K Jul  2 14:32 spack.yaml
```

The closing of the PR [also worked](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16016249990/job/45183353903#step:6:20), the environments are removed properly via spack despite the tarred metadata.

aidanheerdegen (approved):
LGTM

Issue #276, opened 2025-05-22
Support general software build and deployment


aidanheerdegen:
### Introduction

There are tools associated with ACCESS models that ACCESS-NRI needs to build and deploy in a trusted and reliable manner on HPC targets. Python tools are generally deployed in `conda` environments, and in some cases non-python tools are also deployed in `conda` environments. There are limitations to `conda` and in some cases where we want stand-alone tools `spack` might be a better fit.

### Proposal

The proposal is to re-use the `build-cd` infrastructure to build and deploy software: similar software would be grouped together in a single environment (`spack.yaml`).

This use case is covered by the existing infrastructure, except for some assumptions about there being a "top-level" SBR.

aidanheerdegen:
There is a test-case here:

https://github.com/ACCESS-NRI/model-tools/pull/1

The `model-tools` repo was created from the model-deployment repo template with modifications to the `spack.yaml` based on this PR

https://github.com/ACCESS-NRI/ACCESS-OM3/pull/45

aidanheerdegen:
The above test case works to build and deploy to pre-release:

```
$ module use /g/data/vk83/prerelease/modules
$ module load fre-nctools/pr1-5
$ echo $PATH
/g/data/vk83/prerelease/apps/spack/0.22/release/linux-rocky8-x86_64/intel-2021.10.0/fre-nctools-git.2024.05_2024.05-i4evfp27rfqcaksu6xh7ck43gbs535fm/bin:/home/502/aph502/code/git/git-subrepo/lib:/apps/git/2.37.3/bin:/apps/netcdf/4.7.3/bin:/apps/ncview/2.1.7/bin:/home/502/aph502/bin:/opt/pbs/default/bin:/opt/nci/bin:/opt/bin:/opt/Modules/v4.3.0/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/opt/pbs/default/bin:.
$ ls /g/data/vk83/prerelease/apps/spack/0.22/release/linux-rocky8-x86_64/intel-2021.10.0/fre-nctools-git.2024.05_2024.05-i4evfp27rfqcaksu6xh7ck43gbs535fm/bin
check_mask           make_hgrid              nc_null_check
combine-ncc          make_land_domain        plevel.sh
combine_restarts     make_quick_mosaic       remap_land
cp_river_vars        make_regional_mosaic    river_regrid
cr_lake_files        make_remap_file         rmv_parallel_rivers
[...]

PR #295, opened 2025-07-09
Injection of Modules over Checks

### LINKED ISSUE #169: Inject Values for `spack.modules.default.tcl.projections`
## Background

The `spack.modules.default.tcl.projections` section is a section that creates a bit of a headache for users. It adds checks that the version in `spack.packages.*.require[0]` section is equivalent (some exceptions for `@git.`-and-`=RHS`-style versions) to `spack.modules.default.tcl.projections`. 

Unfortunately, this means that users often have to write the same thing twice. For example, changing a version of `mom5` means that they must also change the equivalent projection. This trips up people regularly, enough times that I've kind of gone back on my more extreme what-you-see-is-what-you-get stance that we should never have the CI modify the `spack.yaml`. Even this had been chipped away at earlier anyway - we do sneakily modify the `spack.yaml` to give Prereleases the familiar `MODEL/prX-Y` modulefile name that we all know and love. 

## Implementation Details

* Have the CI inject the appropriate projections into the `spack.yaml` before it is sent off to the deployment targets for install. This should be a fairly simple `yq` in-place edit based on the values in `spack.packages`. 
* If a user overrides the projections by setting the `spack.modules.default.tcl.projections`, let them. Do not inject any of the values.

CodeGat:
Closes #169
References #276

## Background

Having to update versions of packages in multiple places (`spack.packages.*.require[0]` and `spack.modules.tcl.default.projections` mostly) has long been a source of errors and annoyance for users. 

Rather than checking user manifests and reporting errors, we will now inject the relevant projections and includes into the manifest, and not override existing user projections. 

As a little bonus, I've also added a module version of the prerelease modifications that are done in CD, too, so it's independently testable. 

## The PR

- **Add modules injection module**
- **Add requirements for injection modules**
- **Add tests for injection projection module, and inputs/requirements**
- **Add prerelease injection module**
- **Add tests for prerelease injection module**
- **deploy-2-start: Run injection.modules and injection.prerelease modules**
- **deploy-1-setup: Remove projection checks**

### Python Unit Tests

I've added tests for all the important functions in the `scripts.spack_manifest.injection.modules` and `scripts.spack_manifest.injection.prerelease` scripts.
 
These can be invoked with `python3 -m pytest` to run the full suite of `build-cd` tests, or `python3 -m pytest tests/scripts/spack_manifest` for just the new ones. 

### Local Module Runthrough

Can be done via the following commands. They will be most representative if the projections section is removed, or using your own manifests. 

```bash
# For testing projection injection into the manifest
python3 -m scripts.spack_manifest.injection.modules --manifest tests/scripts/spack_manifest/injection/inputs/spack.yaml --root-spec access-om2 --packages "mom5 cice5 oasis3-mct libaccessom2"`

# For testing prerelease modifications to the manifest
python3 -m scripts.spack_manifest.injection.prerelease --manifest tests/scripts/spack_manifest/injection/inputs/prerelease.spack.yaml --root-spec access-om2 --version pr12-12
```

### E2E Workflow Tests

For E2E tests, I used an ephemeral branch `169-inject-projections_TEST` that used the same refs for all intra-pipeline workflow calls, and tested in on `ACCESS-TEST`. See https://github.com/ACCESS-NRI/ACCESS-TEST/pull/48. 

#### Inject Projections As Prerelease

In [this run](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16161252034/job/45613264027?pr=48), you can see the [original manifest](https://github.com/ACCESS-NRI/ACCESS-TEST/pull/48/files#diff-e8582e74fa156f4e5729a850e52b24f2fde2d815c2c9c360f88c4cf90db851ab) both [after projections are injected](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16161252034/job/45613264027?pr=48#step:10:20) (done both in Prerelease and Release) and then [after prerelease modifications are made](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16161252034/job/45613264027?pr=48#step:11:19). The injections are also visible in the artifacts generated by the run, see at the bottom of the summary in the metadata artifact - https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16161252034?pr=48

#### Inject Projections Where They Already Exist

I also tested injecting projections where they were already defined, and it succeeded -  see https://github.com/ACCESS-NRI/ACCESS-ESM1.5/pull/38. This could be for the case where users haven't got around to deleting the projections section.


aidanheerdegen:
In the test run the pre-release injection results in the [following change to projections](https://github.com/ACCESS-NRI/ACCESS-TEST/actions/runs/16161252034/job/45613264027#step:11:47):

```diff
         projections:
-          access-test: '{name}/2025.06.001'
+          access-test: '{name}/pr48-7'
           access-test-component: '{name}/main-{hash:7}'
```
So only the main spec is modified. Is that how we're currently doing it? Do we not get module namespace collisions because the components have their spack hash included, so if they're the same then it's all cool?  What about when we delete the environment, do the module files get cleaned up too?

I'm just wondering if we're doing this, maybe we should add the PR namespace to the component as well, e.g.
```yaml
access-test-component: '{name}/pr48-7/main-{hash:7}'
```
Then we could cleanly remove all module files within the `pr48-*` directories.

CodeGat:
Yep, that's how we're currently doing it. And yeah, we don't get collisions because of the spack hash. 

When we delete the environment, the modulefiles are not cleaned up, unfortunately. See https://github.com/ACCESS-NRI/build-cd/issues/155 - still finding a way to clean them up. I've tried a bunch of `spack`s ways to remove the modulefiles (`spack module tcl rm`, etc) but it seems to mess with existing modulefiles as well. Maybe doing a `spack -e ENV module tcl rm`?

And yes it could be good to put the PR namespace into the components - I'll investigate that. If we can do an environment-aware `module tcl rm` we may not need to, though.

aidanheerdegen:
> And yes it could be good to put the PR namespace into the components - I'll investigate that. 

👍 

> If we can do an environment-aware `module tcl rm` we may not need to, though.

True. As long as `spack` knows which environments are using a tcl module file. If there are multiple, and it doesn't know this, then removing for one environment could break another.

The nice thing about using the name-space is that it would avoid above, and if you can't get `spack` to do what you want, then it's simple to remove them manually.

CodeGat:
Should this forced namespacing be the case for projections supplied by the user?

If so, this goes against the rationale of not overriding user projections. 
And if not, we don't really have a way of knowing what projections are specified by the user in the `prerelease` module (since projection injection happens first). .

aidanheerdegen:
> Should this forced namespacing be the case for projections supplied by the user?

I guess not.
 
> If so, this goes against the rationale of not overriding user projections. And if not, we don't really have a way of knowing what projections are specified by the user in the `prerelease` module (since projection injection happens first). .

It's going to be vanishingly rare for people to bother specifying their own projections. So in that case we live with there being some funky modules sitting around I suppose. Can always clean up anything older than, say, 12 months.

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+        type=str,
+        required=True,
+        help="Name of the root spec of the deployment",
+    )
+
+    parser.add_argument(
+        "--packages",
+        type=str,
+        required=True,
+        help="List of space-separated packages (excluding the root spec) to be considered for projection injection",
```
aidanheerdegen: This will require quoting on the command line to pass more than one package.

It would be good to have some tests that demonstrate acceptable use cases. It is possible to do if the `parse_args` function takes an `args` argument. Then that argument can be passed in the test suite. This also requires the call to `parse_args` in `main` to pass `sys.argv` as an argument.

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+    # Generate the new projections based on the manifest content
+    new_projections: dict[str, str] = generate_projections(
+        manifest, root_spec, defined_projections, projections_to_generate
+    )
+
+    # Inject the new projections into the manifest content, and sort them
+    injected_manifest: dict[str, Any] = manifest
+    injected_manifest["spack"]["modules"]["default"]["tcl"][
+        "projections"
+    ] = new_projections
```
aidanheerdegen: This throws a `KeyError` exception when the `modules` section is not present
```
$ python -m scripts.spack_manifest.injection.projections --root-spec access-om2 --manifest test.yaml --packages "mom5 cice5 netcdf-c"
Extracted version 2023.11.09 from package mom5@git.2023.11.09=access-om2
Extracted version 2023.10.19 from package cice5@git.2023.10.19=access-om2
Extracted version 4.7.4 from package netcdf-c@4.7.4
Extracted version 2024.03.0 from root spec definition access-om2@git.2024.03.0=latest
Traceback (most recent call last):
  File "/Users/aidan/mambaforge/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/aidan/mambaforge/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 236, in <module>
    main()
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 213, in main
    injected_manifest: dict[str, Any] = inject_projections(
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 164, in inject_projections
    injected_manifest["spack"]["modules"]["default"]["tcl"][
KeyError: 'modules'
```

This is a use-case we want to support I think. So a user need not worry about environment modules _at all_.

aidanheerdegen: This might be a use-case for [defaultdict](https://docs.python.org/3/library/collections.html#defaultdict-objects).

aidanheerdegen: Also, we're not modifying `spack.modules.default.tcl.include` at all. If we support the use case of there being no `modules` section then we'd need to add the `spec` and `packages` to `include`.

Don't we want to routinely add the projection packages to `include` in any case?

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+        root_spec_set | (packages & packages_with_versions_defined)
+    ) - defined_projections_set
+
+    # Generate the new projections based on the manifest content
+    new_projections: dict[str, str] = generate_projections(
+        manifest, root_spec, defined_projections, projections_to_generate
+    )
+
+    # Inject the new projections into the manifest content, and sort them
+    injected_manifest: dict[str, Any] = manifest
```
aidanheerdegen: Is this is a shallow copy, a copy by reference? It might be fine if the original `manifest` object isn't modified. Just wanted to point it out.

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+    # Then check if the spec if defined in the traditional, single-target format
+    if not root_spec_definition:
+        root_spec_definition = manifest.get("spack", {}).get("specs", [])[0]
+
+    # If it isn't in either of those places, we don't know where it is
+    if not root_spec_definition:
+        raise ValueError(
+            f"Could not find root spec definition for projection {projection} in the manifest. The spack.yaml is invalid"
+        )
+
```
aidanheerdegen: I don't think this logic works.

When I specified a bogus `root-spec` it didn't error here, but rather in the match logic below:
```
$ python -m scripts.spack_manifest.injection.projections --root-spec access-om --manifest test.yaml --packages "mom5 cice5 netcdf-c access-om2"
Extracted version 4.7.4 from package netcdf-c@4.7.4
Traceback (most recent call last):
  File "/Users/aidan/mambaforge/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/aidan/mambaforge/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 236, in <module>
    main()
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 213, in main
    injected_manifest: dict[str, Any] = inject_projections(
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 158, in inject_projections
    new_projections: dict[str, str] = generate_projections(
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 48, in generate_projections
    projection_version: str = _generate_projection_version_from_spec_or_raise(
  File "/Users/aidan/code/build-cd/scripts/spack_manifest/injection/projections.py", line 121, in _generate_projection_version_from_spec_or_raise
    raise ValueError(
ValueError: Could not extract version from root spec definition access-om2@git.2024.03.0=latest for projection access-om. The root spec needs a version defined.
```
That is because at this point `root_spec_definition` **is** defined.

For this to work I think the version regex below would have to not specify the projection, capture that value in the regex group and compare it to the `root_spec_definition` argument.

CodeGat:
Hey @aidanheerdegen, I've updated this PR to namespace all non-root-spec projections (including user defined after an offline conversation) to be of the form `{name}/prX-Y/VERSION` - changes are in https://github.com/ACCESS-NRI/build-cd/pull/295/files/0de367caa683d3e83202e0daf4c66c54c3589a99..5d0c14fc8c159cd31603574646c989564a38299c

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+    defined_projections_set: set[str] = set(defined_projections.keys())
+    # Get packages that have versions defined in the manifest - we can only generate projections for things with an explicit version
+    packages_with_versions_defined: set[str] = _get_packages_with_versions_defined(
+        manifest
+    )
+
+    # Essentially...we want to generate projections for all packages that don't already have them, provided they have a version defined, or they are the root spec.
+    projections_to_generate: set[str] = (
+        root_spec_set | (packages & packages_with_versions_defined)
+    ) - defined_projections_set
```
aidanheerdegen: This is really elegant. I like it. 

However .. if I pass nonsense/missing packages on the command line they are silently ignored.

Could we check if there are packages that are defined but not present in the manifest and warn a user? I guess it's unlikely in the CI workflow, but might happen if this module is used in other contexts.

scripts/spack_manifest/injection/projections.py
```diff
@@ -0,0 +1,236 @@
...
+    #   packages:
+    #     package1:
+    #       require:
+    #         - "@git.1.0.0"
+    packages: dict[str, Any] = manifest.get("spack", {}).get("packages", {})
+    packages_with_versions_defined: set[str] = set()
+
+    for package_name, package_spec in packages.items():
+        if package_spec["require"][0].startswith("@"):
+            packages_with_versions_defined.add(package_name)
```
aidanheerdegen: We're assuming the version requirement is the first line. Is that something we enforce, i.e. can we rely on this?

aidanheerdegen (commented):
I'll submit this review (half finished) now so I can pull in the most recent commits and I don't get confused between what I've done already and what is new.

scripts/spack_manifest/injection/projections.py
CodeGat: Yep, it's enforced in our spack.yaml schema I believe

scripts/spack_manifest/injection/projections.py
CodeGat: > some tests that demonstrate acceptable use cases

As in, tests for the function `parse_args`?

CodeGat: In any case, I've written some here: https://github.com/ACCESS-NRI/build-cd/pull/295/commits/b034315ad91f7de53da61432f7fa01fe595f6e6e

scripts/spack_manifest/injection/projections.py
CodeGat: Looks like it's a copy-by-reference for those nested structures. Whoops.

scripts/spack_manifest/injection/projections.py
CodeGat: Could it be valid to have a package defined in `--packages` but have no entry in `spack.packages.*`? Like, we want to keep track of this package but don't have any requirements/constraints to place on it?

scripts/spack_manifest/injection/projections.py
CodeGat: This is done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/2b5139f76f86835e0150984bee33f79256ec8d76

scripts/spack_manifest/injection/projections.py
CodeGat: I've now wrapped those in `dict(manifest)` so it is a true copy

scripts/spack_manifest/injection/projections.py
CodeGat: Updated the logic in this now-renamed function: https://github.com/ACCESS-NRI/build-cd/pull/295/files#diff-285e75d0870eac01b757c96c593018d0170b3896fee2bfb616b2a7abef77fb76R159-R201

scripts/spack_manifest/injection/projections.py
CodeGat: Missing packages would be a bit difficult to find since they could be a dependency that isn't explicitly specified in the abstract manifest, but only show in the concretized version

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+def main():
+    # Get inputs
+    args = parse_args(sys.argv[1:])
+
+    packages: set[str] = set(args.packages.split())
+
+    with open(args.manifest, "r") as file:
+        manifest: dict[str, Any] = yaml.safe_load(file)
+
+    # Inject manifeest with projections and includes
```
aidanheerdegen:
```suggestion
    # Inject manifest with projections and includes
```
Typo

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+    if args.output:
+        with open(args.output, "w") as output_file:
+            yaml.dump(
+                manifest_with_projections_and_includes,
+                output_file,
+                default_flow_style=False,
+                sort_keys=False,
+            )
```
aidanheerdegen: Another option would be to save the dump to a variable which you then `print` and optionally `write` to a file.

It's honestly not going to make much difference in speed, but is a bit more "programmy".

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Get projections that are already defined in the manifest - we don't want to redefine these
+    defined_projections: dict[str, str] = _get_defined_projections(manifest)
+    defined_projections_set: set[str] = set(defined_projections.keys())
+    # Get packages that have versions defined in the manifest - we can only generate projections for things with an explicit version
+    packages_with_versions_defined: set[str] = _get_packages_with_versions_defined(
+        manifest
+    )
+
+    # Essentially...we want to generate projections for all packages that don't already have them, provided they have a version defined
```
aidanheerdegen:
```suggestion
    # Generate projections for all packages that don't already have them, provided they have a version defined
```
I'm a fan of relaxed conversational chat, but in this case I think it obscures meaning, and we should always have in mind those for whom English is not their primary language and keep it as clear as possible.

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+    manifest: dict[str, Any], root_spec: str, packages: set[str]
+) -> dict[str, Any]:
+    # We want to inject the includes for the root spec, all packages defined in --packages, and existing includes in the manifest.
+    existing_includes: set[str] = _get_defined_includes(manifest)
+
+    # Includes are the union of packages, and existing includes - we add the root spec later
+    includes: set[str] = packages | existing_includes
+
+    # To sort, we want to ensure that the root spec is always first in the list, and the rest are sorted alphabetically
+    sorted_include: list[str] = [root_spec] + sorted(includes)
```
aidanheerdegen:
```suggestion
sorted_include: list[str] = [root_spec] + sorted(includes - set([root_spec]))
```
If `root_spec` already exists in `existing_includes` then it will appear twice in `sorted_includes`. The suggestion above removes it in that case.

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Finally, we inject the includes into and updated manifest, which we return
```
aidanheerdegen:
```suggestion
    # Finally, inject the includes into a new copy of the manifest, which is returned
```
Just stressing that it is a new, deep, copy of the manifest

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+        root_spec_definition = manifest.get("spack", {}).get("specs", [])[0]
+
+    # If it isn't in either of those places, we don't know where it is
+    if not root_spec_definition:
+        raise ValueError(
+            f"Could not find root spec definition for root spec {root_spec_name} in the manifest. The spack.yaml is invalid"
+        )
+
+    # Now we can extract the actual version from the root spec definition
+    version_regex = re.compile(r"([^@]+)@(?:git.)?([^+~= ]+).*")
```
aidanheerdegen: We have four different places where regex is used for extracting versions from specs. This one, the one below (which is simpler for packages), one in `remove_potential_root_spec_git_version` and a function `get_ref_from_spack_spec_version_or_raise` in `release_provenance/tracking_services_data.py`.

Is there potential for all to use a single (well tested) function or class? 

For a class something analogous to `pathlib.Path` with member functions to extract spec name, version, variants etc might be worthwhile at this point. This might be something we make an issue for a implement at a later date.

scripts/spack_manifest/injection/modules.py
```diff
@@ -0,0 +1,279 @@
...
+        help="Path to the output file where the modified manifest will be saved",
+    )
+
+    parsed_args = parser.parse_args(args)
+
+    # Verifying that --packages are space-separated, which is a bit different from the usual comma-separated lists
+    if ',' in parsed_args.packages:
+        raise ValueError(
+            "The --packages argument must be a space-separated list of package names."
+        )
```
aidanheerdegen: If you have this check it would be good to test that it works as expected. Maybe there is a test and I missed it, in which case please ignore.

aidanheerdegen (commented):
Looks really good, thanks. I do have some suggestions still though.

scripts/spack_manifest/injection/projections.py
aidanheerdegen: > Could it be valid to have a package defined in --packages but have no entry in spack.packages.*? Like, we want to keep track of this package but don't have any requirements/constraints to place on it?

Just for completeness, we discussed IRL and decided we would not support generating packages that did not have a version specified in the packages section. This has been implemented I believe, so will resolve this.

scripts/spack_manifest/injection/modules.py
CodeGat: Good point, done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/6d575f33dd581ed821cc83f172cd21aabef0057f

scripts/spack_manifest/injection/modules.py
CodeGat: Yep, it's tested here: https://github.com/ACCESS-NRI/build-cd/blob/6d575f33dd581ed821cc83f172cd21aabef0057f/tests/scripts/spack_manifest/injection/test_modules.py#L227-L240

scripts/spack_manifest/injection/modules.py
CodeGat: Good point, done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/15395519c850cad952765d835da0a40c20b92c1a

scripts/spack_manifest/injection/modules.py
CodeGat: Good point - duplication bad. I'll get to it now.

CodeGat: Whew, this took a while. The changes required are in https://github.com/ACCESS-NRI/build-cd/pull/295/files/587c536645273fc5f10129e8fb0b5795f9fee728..68828df22a41c2618549d7200eebef11bf77d0b3

scripts/spack_manifest/getter.py
```diff
@@ -0,0 +1,223 @@
...
+        defs: list[dict[str, Any]] = self.manifest.get("spack", {}).get("definitions", [])
+        specs: list[str] = self.manifest.get("spack", {}).get("specs", [])
+
+        # It's either in the multi-target format or the single target format, we just need to find which
+            # The multi-target format is of the form:
+        # spack:
+        #   definitions:
+        #     - ROOT_PACKAGE: [access-om2@git.2025.05]
+        #     # ...
+        root_package_def = next((d["ROOT_PACKAGE"] for d in defs if "ROOT_PACKAGE" in d), [])
```
aidanheerdegen: Fancy schmancy python!

scripts/spack_manifest/getter.py
```diff
@@ -0,0 +1,223 @@
...
+    # def get_compiler_name(self) -> str:
+    #     pass
+
+    # def get_compiler_version(self) -> str:
+    #     pass
+
+    ## Get architecture information
+
+    # def get_arch(self) -> str:
+    #     pass
```
aidanheerdegen: Is the intention to provide these methods at a later date?

If so, would it be better to not have them commented, but instead return a `NotImplemented` exception?

scripts/spack_manifest/getter.py
```diff
@@ -0,0 +1,223 @@
...
+            raise NoSectionComponentError("Root spec component 'name' could not be extracted from the root spec string.")
+
+        return match.group(0)
+
+    ## Get version-related information
+
+    # def get_full_version(self) -> str:
+    #     pass
+
+    def get_ref(self) -> str:
```
aidanheerdegen, line 79: Would be good to have a comment with an example or two of specs it would match against

scripts/spack_manifest/getter.py
```diff
@@ -0,0 +1,223 @@
...
+
+        return self.packages[name].get("require", [])
+
+    def get_package_ref_requirement(self, name: str) -> str:
+        requirements = self.get_package_requirements(name)
+
+        if requirements == []:
+            raise NoSectionComponentError(f"Package component 'version' could not be extracted from the package requirements string for '{name}'.")
+
+        match = re.match(r"@(?:git\.)?([^+~=% ]+)", requirements[0])
```
aidanheerdegen, line 213: Would be good to have a comment illustrating the uses cases and the expected match.

scripts/spack_manifest/injection/modules.py
```diff
@@ -47,22 +52,31 @@ def inject_projections(
 ) -> dict[str, Any]:
 
     # Get projections that are already defined in the manifest - we don't want to redefine these
-    defined_projections: dict[str, str] = _get_defined_projections(manifest)
-    defined_projections_set: set[str] = set(defined_projections.keys())
+    # defined_projections: dict[str, str] = _get_defined_projections(manifest)
+    # defined_projections_set: set[str] = set(defined_projections.keys())
```
aidanheerdegen: I guess the intention is to remove the commented code?

scripts/spack_manifest/getter.py
```diff
@@ -0,0 +1,223 @@
...
+    def from_file(cls, manifest_path: str) -> 'Includes':
+        from yaml import safe_load
+
+        with open(manifest_path, "r") as file:
+            manifest = safe_load(file)
+
+        return cls(manifest)
+
+    def get(self) -> list[str]:
+        return self.manifest.get("spack", {}).get("modules", {}).get("default", {}).get("tcl", {}).get("include", [])
```
aidanheerdegen: IWould it make sense to have a `.set()` method as well, and use that instead of directly setting the includes at the end of `inject_projections`?

scripts/spack_manifest/injection/prerelease.py
```diff
@@ -79,28 +81,21 @@ def add_namespace_to_other_projection_versions(
...
+    if root_spec_from_manifest.has_git_ref():
         # Remove the @git version and then add later contraints back
-        spec_without_version = " ".join(m.lstrip() for m in component_match.groups())
-        manifest["spack"]["specs"][0] = spec_without_version.strip()
+        manifest["spack"]["specs"][0] = f"{name} {constraints}".strip()
     else:
-        raise ValueError(
-            f"Root spec '{root_spec}' not found in the manifest or does not match expected format."
-        )
+        print(f"The root spec '{name}' does not have a git ref, so no changes are made.")
```
aidanheerdegen: This is really nice and clean and easier to read and parse. 👍

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                    {
+                        "ROOT_PACKAGE": ["access-om2@git.1.0.0"]
+                    }
+                ]
+            }
+        }
+
+        root_spec_getter = RootSpec(manifest)
+
+        assert root_spec_getter.root_spec == "access-om2@git.1.0.0", "Root spec should be correctly initialized from multi-target manifest."
```
aidanheerdegen: I know it isn't necessary for the test, but without a `ROOT_SPEC` matrix it is hard to interpret how this test is supposed to work. You don't have to include one, but a comment that the missing matrix is what connects `ROOT_PACKAGE` to `$ROOT_SPEC` would be welcome.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        root_spec_getter = RootSpec(manifest)
+        with pytest.raises(NoSectionComponentError):
+            root_spec_getter.get_name()
+
+    def test_get_ref__valid_git(self):
+        manifest = {
+            "spack": {
+                "specs": ["access-om2@git.1.0.0"]
```
aidanheerdegen: Might be worth putting in some examples of other supported version types, like git hashes.

Could use `pytest` paramterization (sic) to easily specify a few different supported version types.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+
+        root_spec_getter = RootSpec(manifest)
+        ref = root_spec_getter.get_ref()
+
+        assert ref == "1.0.0", "Root spec ref should be correctly extracted."
+
+    def test_get_ref__valid_non_git(self):
+        manifest = {
+            "spack": {
+                "specs": ["access-om2@1.0.0"]
```
aidanheerdegen: We don't impose any idea of what we think a version should look right? I think that's ok, but I wonder about people accidentally using a git ref with the `@git.` prefix.

aidanheerdegen: I should say I don't know we should do anything about it ...

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+        manifest = {
+            "spack": {
+                "specs": ["access-om2@git.1.0.0=develop %compiler@2.0.0 +debug"]
+            }
+        }
+
+        root_spec_getter = RootSpec(manifest)
+        ref = root_spec_getter.get_ref()
+
+        assert ref == "1.0.0", "Root spec ref should be correctly extracted with Spack version and constraints."
```
aidanheerdegen: This could be collapsed into a single parameterized test but that may be at the loss of a fine-grained error message.

If you did go that way, would be nice to have a version of the tests that collapsed the whitespace and confirmed they worked as expected.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+            root_spec_getter.get_ref()
+
+    @pytest.mark.parametrize(
+        "root_spec",
+        [
+            "",
+            "%compiler@2.0.0",
+            " %compiler@2.0.0 +debug",
+            "+debug",
+            " ~debug",
```
aidanheerdegen: Is the whitespace required for `~debug`? If not it should be straightforward to add a couple of extra tests, e.g.

```suggestion
            "+debug",
            "~debug",
            " +debug",
            " ~debug",
```
Or perhaps these are effectively redundant?

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                    "package2": {"require": ["@git.2.0.0"]},
+                }
+            }
+        }
+
+        packages_getter = Packages(manifest)
+
+        assert packages_getter.packages == {
+            "package1": {"require": ["@git.1.0.0"]},
+            "package2": {"require": ["@git.2.0.0"]},
```
aidanheerdegen: Would it be considered good practice to just compare to `manifest['spack']['packages']`? It makes it less verbose and easier to maintain.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+    def test___init___invalid_no_packages(self):
+        manifest = {
+            "spack": {
+                "packages": {}
+            }
+        }
+
+        packages_getter = Packages(manifest)
+
+        assert packages_getter.packages == {}, "Packages should be initialized as an empty dictionary if no packages are defined."
```
aidanheerdegen: So we require an empty packages section? It isn't valid to have `packages` missing altogether?

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def test___init___invalid_no_packages(self):
+        manifest = {"spack": {}}
+
+        with pytest.raises(NoSectionError):
+            Packages(manifest)
+
+    def test_from_file__valid(self):
+        manifest_path = "tests/scripts/spack_manifest/injection/inputs/prerelease.spack.yaml"
+        packages_getter = Packages.from_file(manifest_path)
```
aidanheerdegen: This isn't testing it has been read correctly. This could be done by defining a manifest, writing it to a file and then checking the instantiation from file is done correctly.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                    "package1": {"require": ["@git.1.0.0"]},
+                    "package2": {"require": ["@git.2.0.0"]},
+                }
+            }
+        }
+
+        packages_getter = Packages(manifest)
+        requirements = packages_getter.get_package_requirements("package1")
+
+        assert requirements == {"require": ["@git.1.0.0"]}, "Package requirements should be correctly retrieved."
```
aidanheerdegen: Could test both requirements with a parameterised version of the test.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def test_get_package_requirements__valid(self):
+        manifest = {
+            "spack": {
+                "packages": {
+                    "package1": {"require": ["@git.1.0.0"]},
+                    "package2": {"require": ["@git.2.0.0"]},
+                }
+            }
+        }
```
aidanheerdegen: This manifest is reused a lot. Could make it a fixture.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+    def test_get_package_requirements__invalid_no_package(self):
+        manifest = {
+            "spack": {
+                "packages": {}
+            }
+        }
+
+        packages_getter = Packages(manifest)
+        with pytest.raises(NoSectionComponentError):
+            packages_getter.get_package_requirements("nonexistent_package")
```
aidanheerdegen, line 274: I think the test would be better if there were some packages defined. Would be straightforward if manifest was a fixture as suggested above.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+    def test_get_package_full_version_requirement__invalid_no_package(self):
+        manifest = {
+            "spack": {
+                "packages": {}
+            }
+        }
+
+        packages_getter = Packages(manifest)
+        with pytest.raises(NoSectionError):
+            packages_getter.get_package_full_version_requirement("nonexistent_package")
```
aidanheerdegen, line 299: As above.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        expected = {"package1": "package1/1.0.0", "package2": "package2/2.0.0"}
+
+        projections_getter = Projections(manifest)
+        result = projections_getter.get()
```
aidanheerdegen: Just from a lines of code point of view could collapse with no loss of clarity

```suggestion
        result = Projections(manifest).get()
```

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+            result == expected
+        ), "A valid manifest should return all defined projections."
+
+    def test__get_defined_projections__no_projections(self):
+        manifest = {"spack": {"modules": {"default": {"tcl": {"projections": {}}}}}}
+
+        expected = dict()
+
+        projections_getter = Projections(manifest)
+        result = projections_getter.get()
```
aidanheerdegen:
```suggestion
        result = Projections(manifest).get()
```

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+            result == expected
+        ), "Manifest without projections should return an empty set."
+
+    def test__get_defined_projections__no_modules_section(self):
+        manifest = {"spack": {}}
+
+        expected = dict()
+
+        projections_getter = Projections(manifest)
+        result = projections_getter.get()
```
aidanheerdegen:
```suggestion
        result = Projections(manifest).get()
```

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        expected = ["root-spec", "package2", "package3"]
+
+        includes_getter = Includes(manifest)
+        result = includes_getter.get()
```
aidanheerdegen:
```suggestion
        result = Includes(manifest).get()
```

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+                    "default": {
+                        "tcl": {
+                            "include": ["root-spec", "package2", "package3"]
+                        }
+                    }
+                }
+            }
+        }
+
+        expected = ["root-spec", "package2", "package3"]
```
aidanheerdegen: As above, could just compare directly to `manifest['spack']['modules']['default']['tcl']['include']`, or defined `expected` first and use that to populate `manifest`.

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+            result == expected
+        ), "A valid manifest should return all defined includes."
+
+    def test__get_defined_includes__no_includes(self):
+        manifest = {"spack": {"modules": {"default": {"tcl": {"include": []}}}}}
+
+        expected = []
+
+        includes_getter = Includes(manifest)
+        result = includes_getter.get()
```
aidanheerdegen:
```suggestion
        result = Includes(manifest).get()
```

tests/scripts/spack_manifest/test_getter.py
```diff
@@ -0,0 +1,480 @@
...
+            result == expected
+        ), "Manifest without includes should return an empty set."
+
+    def test__get_defined_includes__no_modules_section(self):
+        manifest = {"spack": {}}
+
+        expected = []
+
+        includes_getter = Includes(manifest)
+        result = includes_getter.get()
```
aidanheerdegen:
```suggestion
        result = Includes(manifest).get()
```

aidanheerdegen (changes requested):
I have reviewed some ... still some more to look through.

Edit: not sure actually, so I'll leave it as-is and let me know if you need another review.

scripts/spack_manifest/injection/modules.py
aidanheerdegen: Thanks. Makes for clearer and easier to read code.

scripts/spack_manifest/getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/9ac842de23cc1166e8584ce604b976f8dc08d77b

scripts/spack_manifest/getter.py
CodeGat: Also done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/9ac842de23cc1166e8584ce604b976f8dc08d77b

scripts/spack_manifest/getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/9ac842de23cc1166e8584ce604b976f8dc08d77b

scripts/spack_manifest/injection/modules.py
CodeGat: https://github.com/ACCESS-NRI/build-cd/pull/295/commits/bd619385866f0b5d86144b5b5a62e87947afefd5

tests/scripts/spack_manifest/test_getter.py
CodeGat: Fair enough, added a comment to that effect: https://github.com/ACCESS-NRI/build-cd/pull/295/commits/6375b8ad69be57c3b54a0bc0be7524c74a184902

tests/scripts/spack_manifest/test_getter.py
CodeGat: We do impose the `@git.` part in the schema, but in the case where users are using this module outside of CI without validating their manifest, it will be picked up here.

tests/scripts/spack_manifest/test_getter.py
CodeGat: Have made heavy use of parameterization in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/804e577a4d5b6e7a4f881a2cff63ab0dc9a0e96f

tests/scripts/spack_manifest/test_getter.py
CodeGat: Have parameterized tests in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/804e577a4d5b6e7a4f881a2cff63ab0dc9a0e96f

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/804e577a4d5b6e7a4f881a2cff63ab0dc9a0e96f

tests/scripts/spack_manifest/test_getter.py
CodeGat: Fair enough, done via https://github.com/ACCESS-NRI/build-cd/pull/295/commits/804e577a4d5b6e7a4f881a2cff63ab0dc9a0e96f

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/cac77e8f06ec5cd991abb3fe2e7f02698d0a5384

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/cac77e8f06ec5cd991abb3fe2e7f02698d0a5384

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/cac77e8f06ec5cd991abb3fe2e7f02698d0a5384

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/a00418c034d20b66fd88775da833e6164d63b710

tests/scripts/spack_manifest/test_getter.py
CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/804e577a4d5b6e7a4f881a2cff63ab0dc9a0e96f

tests/scripts/spack_manifest/test_getter.py
CodeGat: Whoops, test should be `__valid` not invalid! I'm parameterized this test as well in https://github.com/ACCESS-NRI/build-cd/pull/295/commits/1a0f1b8fac8b436637996bc1df2718e28bf5d3d3

scripts/spack_manifest/getter.py
CodeGat: Hmm, yes that's a good point. Might have to change the name of the class and tests, if we add as set method to the getter class :laughing:

aidanheerdegen: No worries. Leave for later on.

CodeGat:
There is potential to need to move `vars.BUILD_DB_PACKAGES` into a file rather than a `var`

tests/scripts/spack_manifest/test_getter.py
aidanheerdegen: Thanks. It makes it easier to see the various combinations in the parameterised form.

aidanheerdegen (approved):
LGTM

PR #308, opened 2025-08-01
Tar up prerelease installations on success or failure

### LINKED ISSUE #307: Also tar up installs that fail
Currently, we tar up installations in spack once they succeed. But it turns out that failed installs don't clean up after themselves environment-wise, either. 

In order to save inodes, we should tar up environments even if they fail to install.

CodeGat:
Closes #307

## Background

We tar up environment metadata only on success, which can lead to many environments not being tarred up. 

## The PR

* Attempt to do the tar step on success or failure of the `deploy` step.

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -275,7 +275,11 @@ jobs:
           rm ./root-spec-pkg-hash.${{ github.run_id }}
 
       - name: Tar spack environment metadata
-        if: inputs.deployment-type == 'Prerelease'
+        # We want to attempt to tar up the .spack-env directory even if the install fails, so we can save inodes.
+        if: >-
+          always() &&
+          (steps.deploy.conclusion == 'success' || steps.deploy.conclusion == 'failure') &&
+          inputs.deployment-type == 'Prerelease'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -275,7 +275,11 @@ jobs:
           rm ./root-spec-pkg-hash.${{ github.run_id }}
 
       - name: Tar spack environment metadata
-        if: inputs.deployment-type == 'Prerelease'
+        # We want to attempt to tar up the .spack-env directory even if the install fails, so we can save inodes.
+        if: >-
+          always() &&
+          (steps.deploy.conclusion == 'success' || steps.deploy.conclusion == 'failure') &&
```
aidanheerdegen, line 281: I assume you wouldn't have done it unless it was necessary, but you do still need to check the status of `steps.deploy.conclusion` when with `always()`?

aidanheerdegen (approved):
LGTM

.github/workflows/deploy-2-start.yml
CodeGat: You do, because the other two statuses are 'skipped' and 'cancelled', in which case we just want to get to the end of the job quickly

PR #319, opened 2025-09-01
Pass in optional TRACKING_SERVICES_REPO_FETCH_TOKEN to upload script

### LINKED ISSUE #282: Release Database Fails to Get Repo Info From Private Repository
## Background

When attempting to validate the ref of a dependency spack packages private reposiory, the `tracking_services_data.py` notes that the ref doesn't exist, see https://github.com/ACCESS-NRI/ACCESS-ESM1.6/actions/runs/15549058082/job/43777363256#step:7:33.

We are currently using the default token given by GitHub Actions, which doesn't allow access to other, private repositories. 

## Implementation details

* Should we instead require env to be set that is the required token?

CodeGat:
Closes #282

## Background

The upload script fails for Model components with private repositories (like `ACCESS-ESM1.*`), which means they need to be uploaded manually. 

The fix for this is to be able to pass in a PAT that has those repos readable. 

## The PR

* Allow passing in of an optional `TRACKING_SERVICES_REPO_FETCH_TOKEN` to the upload script

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -309,6 +309,7 @@ jobs:
         env:
           TRACKING_SERVICES_POST_URL: ${{ vars.TRACKING_SERVICES_POST_URL }}
           TRACKING_SERVICES_POST_TOKEN: ${{ secrets.TRACKING_SERVICES_POST_TOKEN }}
+          TRACKING_SERVICES_REPO_FETCH_TOKEN: ${{ secrets.TRACKING_SERVICES_REPO_FETCH_TOKEN }}
         run: |
           python -m scripts.release_provenance.tracking_services_data \
             --repository ${{ github.repository }} \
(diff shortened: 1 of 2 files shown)


jo-basevi (dismissed):
Looks good to me

.github/workflows/cd.yml
```diff
@@ -309,6 +309,7 @@ jobs:
         env:
           TRACKING_SERVICES_POST_URL: ${{ vars.TRACKING_SERVICES_POST_URL }}
           TRACKING_SERVICES_POST_TOKEN: ${{ secrets.TRACKING_SERVICES_POST_TOKEN }}
+          TRACKING_SERVICES_REPO_FETCH_TOKEN: ${{ secrets.TRACKING_SERVICES_REPO_FETCH_TOKEN }}
```
jo-basevi, line 312: Is this an empty string if not defined in secrets?

jo-basevi: If so, then does the check in the python script will need to check if `$TRACKING_SERVICES_REPO_FETCH_TOKEN` is not an empty string

CodeGat: Yep! Although good point...I'll do an empty string check as well

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/319/commits/6723f1cc52f9da554e368352f5ca6d7d9297a782

jo-basevi (approved):
Thanks! Looks good to me

PR #331, opened 2025-11-25
Support Custom Root Spec Modulefiles in Prerelease

### LINKED ISSUE #329: Custom Module Names
To allow install of multiple spack top-level specs within one deployment, we need build-cd to support customisation of module names.

Currently, customised module names are overridden by the pre-release module injection part of build-cd.

For example, access-om2 requires multiple builds of cice5 for each resolution in the resulting configurations.

So, there may be a spec (in the future) like this:

```
  specs:
    - access-om2@git.2025.12.000 ^cice5 nxglob=360 nyglob=300 blckx=15 blcky=300 mxblcks=1
    - access-om2 ^cice5 nxglob=1440 nyglob=1080 blckx=30 blcky=27 mxblcks=4
    - access-om2 ^cice5 nxglob=3600 nyglob=2700 blckx=18 blcky=15 mxblcks=22
 ```

which can be interpreted as three installs, one for the 1degree configuration, one for a 0.25 degree and one for 0.01 degree.

Spack supports setting module names like this:

```
    default:
      tcl:
        projections:
          access-om2: '{name}/2025.12.000-{^cice5.variants.nxglob.value}x{^cice5.variants.nyglob.value}'
```

And the "Modules Injection" looks fine, [doesn't change](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/19558860065/job/56007080701#step:11:74) this line

However "Prerelease Injection" then [overrides](https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/19558860065/job/56007080701#step:12:74) it:

```
access-om2: '{name}/pr125-9'
```

One good solution would be that the _pr_ number gets inserted after the first `/`, rather than overriding the name, so the resulting module name would be:

```
access-om2: '{name}/pr125-9/2025.12.000-{^cice5.variants.nxglob.value}x{^cice5.variants.nyglob.value}'
```

(I guess if the are no _/_, then append it to the end)

### PR DESCRIPTION
**Pull request by CodeGat**
Closes #329

## Background

A corner case in our Prerelease Injection is in the case where a custom root spec projection is overwritten by the prerelease injection. This is usually fine, as we don't put anything too important in the projections. 

However, in the case where we have multiple top-level specs that share the same name, but different properties (eg. variants in dependencies that need to added to the modulefile, like in `{name}/2025.11.000-{^cice5.variants.x}x{^cice5.variants.y}`), we overwrite it to `{name}/prX-Y`. 

In this PR, rather than directly overwriting the custom projection, we infix the PR number, so we get `{name}/prX-Y/VERSION`, eg. `{name}/pr12-23/2025.11.000-{^cice.variants.x}x{^cice.variants.y}`. This gives us both modulefile uniqueness, and capture the unique properties of the custom projection

## The PR

In this PR:

- **deploy-2-start: infix prerelease version into existing projection**
- **Update tests**

## Testing

Rerun and updated unit tests.

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -137,13 +137,9 @@ jobs:
         working-directory: cd
         run: |
           original_root_spec_projection=$(yq '.spack.modules.default.tcl.projections."${{ steps.manifest.outputs.root-spec-name }}"' ../${{ inputs.spack-manifest-path }})
-          if [[ "$original_root_spec_projection" == "null" ]]; then
-            # If there is no custom projection for the root spec, we just use the version as the projection, of the form `{name}/prX-Y`
-            echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
-          else
-            # For custom root spec projections, the prerelease portion will be infixed
-            # Eg. `{name}/2025.01.000-{^dep.variants.x}` -> `{name}/prX-Y/2025.01.000-{^dep.variants.x}`
-            echo "version=${{ inputs.version }}/${original_root_spec_projection#*/}" >> $GITHUB_OUTPUT
+          if [[ "$original_root_spec_projection" != "null" ]]; then
+            # For custom root spec projections, we want to take everything after the first `/`
+            echo "custom-projection=${original_root_spec_projection#*/}" >> $GITHUB_OUTPUT
           fi
 
           python -m scripts.spack_manifest.injection.modules \
@@ -161,6 +157,7 @@ jobs:
           python -m scripts.spack_manifest.injection.prerelease \
             --manifest ../${{ inputs.spack-manifest-path }} \
             --version ${{ steps.modules-injection.outputs.version }} \
+            ${{ steps.modules-injection.outputs.custom-projection != '' && format('--custom-root-projection {0}', steps.modules-injection.outputs.custom-projection ) || ''}} \
             --spack-packages-path ${{ steps.path.outputs.spack-packages }} \
             --output ../${{ inputs.spack-manifest-path }}
 
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -138,8 +138,7 @@ jobs:
         run: |
           original_root_spec_projection=$(yq '.spack.modules.default.tcl.projections."${{ steps.manifest.outputs.root-spec-name }}"' ../${{ inputs.spack-manifest-path }})
           if [[ "$original_root_spec_projection" != "null" ]]; then
-            # For custom root spec projections, we want to take everything after the first `/`
-            echo "custom-projection=${original_root_spec_projection#*/}" >> $GITHUB_OUTPUT
+            echo "custom-projection=${original_root_spec_projection}" >> $GITHUB_OUTPUT
           fi
 
           python -m scripts.spack_manifest.injection.modules \
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -133,8 +133,19 @@ jobs:
 
       - name: Spack Manifest - Modules Injection
         # Injects relevant module projections and includes into the spack manifest so users don't have to
+        id: modules-injection
+        working-directory: cd
         run: |
-          cd cd
+          original_root_spec_projection=$(yq '.spack.modules.default.tcl.projections."${{ steps.manifest.outputs.root-spec-name }}"' ../${{ inputs.spack-manifest-path }})
+          if [[ "$original_root_spec_projection" == "null" ]]; then
+            # If there is no custom projection for the root spec, we just use the version as the projection, of the form `{name}/prX-Y`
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: scripts/spack_manifest/injection/prerelease.py [hunk 1]
```diff
@@ -129,10 +129,18 @@ def remove_potential_root_spec_git_version(manifest: dict[str, Any]) -> dict[str
 def update_root_spec_projection_version(
     manifest: dict[str, Any], root_spec_name: str, root_spec_version: str
 ) -> dict[str, Any]:
-    updated_version: str = f"{{name}}/{root_spec_version}"
-
     manifest.setdefault("spack", {}).setdefault("modules", {}).setdefault("default", {}).setdefault("tcl", {}).setdefault("projections", {})
 
+    existing_root_spec_projection: str | None = manifest["spack"]["modules"]["default"]["tcl"]["projections"].get(root_spec_name, None)
+
+    if existing_root_spec_projection:
```
jo-basevi: Is a "/" always in `existing_root_spec_projection`?

jo-basevi (dismissed):
LGTM

scripts/spack_manifest/injection/prerelease.py
CodeGat: Yep. In the injection script that runs before the `prerelease.py` one, we add one of the form `{name}/VERSION` if it doesn't already exist. Although...there is a possibility that someone adds a custom projection that is just the name, like `{name}` or something...I'll update it!

CodeGat: I've updated it so it can accomodate projections with no slash, even though that would be a rare occurence, https://github.com/ACCESS-NRI/build-cd/pull/331/commits/c1bc5c7313b7ba6927f25d9b9db50abddade20a6

CodeGat:
@jo-basevi I've updated the solution so it doesn't use the `prerelease.py` bit at all - instead we pass in the bit after `/` (for custom projections) into the `--version` flag.

CodeGat:
@jo-basevi I've further updated the PR, rather than putting the custom bit in the `--version` input, there is now a `--custom-root-projection` arg that is added to the root spec if it is defined.

.github/workflows/deploy-2-start.yml
```diff
@@ -133,8 +133,15 @@ jobs:
...
       - name: Spack Manifest - Modules Injection
         # Injects relevant module projections and includes into the spack manifest so users don't have to
+        id: modules-injection
+        working-directory: cd
         run: |
-          cd cd
+          original_root_spec_projection=$(yq '.spack.modules.default.tcl.projections."${{ steps.manifest.outputs.root-spec-name }}"' ../${{ inputs.spack-manifest-path }})
+          if [[ "$original_root_spec_projection" != "null" ]]; then
+            # For custom root spec projections, we want to take everything after the first `/`
+            echo "custom-projection=${original_root_spec_projection#*/}" >> $GITHUB_OUTPUT
```
jo-basevi: Fairly minor as it'll work for the vast majority of cases, but I wonder if passing the unchanged `original_root_spec_projection` to the python script, then in python it'll be easy to do something like:

```python
prefix = "{name}/"
if custom_root_projection.startswith(prefix):
    custom_root_projection = s[len(prefix):]
```

jo-basevi (dismissed):
Thanks @CodeGat, looks good to me. Feel free to ignore the one comment

.github/workflows/deploy-2-start.yml
CodeGat: Good idea, done in https://github.com/ACCESS-NRI/build-cd/pull/331/commits/3ded8976624d792c1c36d0cba051a9c122282727

scripts/spack_manifest/injection/prerelease.py
```diff
@@ -132,7 +132,12 @@ def update_root_spec_projection_version(
 ) -> dict[str, Any]:
 
     if custom_root_projection is not None and custom_root_projection != "":
-        updated_version: str = f"{{name}}/{root_spec_version}/{custom_root_projection}"
+        projection_components = custom_root_projection.split("/", 1)
+
+        if len(projection_components) == 1:
+            updated_version: str = f"{{name}}/{root_spec_version}/{projection_components[0]}"
```
jo-basevi, line 138: If the `custom_root_projection = "{name}"`, you could have `updated_version="{name}/pr1-1/{name}"`, but maybe that'll be a motivator to not set a custom projection to `{name}`!

Also `{name}-{some-value}/{some-other-value}`  would become `{name}/{root_spec_version}/{some-other-value}`

jo-basevi (approved):
Looks good! Just noted some edge cases that probably will never apply

scripts/spack_manifest/injection/prerelease.py
CodeGat: Good points. I think that the non-`{name}/VERSION` cases are essentially erroneous projections, but in prereleases I think they're ok. Similar to `{name}-BLAH/VERSION`, we should put variations on the build in the VERSION part rather than the module name

PR #357, opened 2026-02-12
Metadata Creation Uses `spack --config-scope` Scopes

### LINKED ISSUE #356: Failed to get metadata from `restricted` install directory
See md5-hashless metadata from run https://github.com/ACCESS-NRI/ACCESS-ESM1.6/actions/runs/21928764599/job/63327858918?pr=168#step:18:55

## Background

We pass in the custom installation directory for UKMO-restricted data through a `spack --config-scope`, not through an environment. This means that the metadata creation script is currently uninformed about the custom restricted directory, which leads to it getting the default prefix for a given package. This package doesn't have to be installed either, and since we don't have the custom scope we need, we get the assumed prefix, which has no binaries, and hence no md5 sums. 

## Implementation Plan

See if there is a way via `spack`s API to load custom configuration scopes for a given environment, or invocation.

CodeGat:
Closes #356

## Background

Currently, if we install a package that uses a custom `spack --config-scope`, the metadata creation logic doesn't know about this, leading to certain models binaries that use restricted directories (like `ACCESS-ESM1.6`) to not be able to be `md5` hashed. 

## The PR

* Update the `get_data_from_deployment.py` script to take a `--config-scopes` arg, which loads the appropriate `spack-config/custom/cd` scopes.

## Testing

Tested via the docker image: `ghcr.io/access-nri/build-ci-upstream:rocky-v1.1-2026.02.000`, in which I copied the script and ran:

```sh
spack env activate --create e
spack install --add access-esm1p6
./get_data_from_deployment.py \
  --environment /opt/environments/e \
  --deployment-name access-esm1p6 \
  --packages um7 \
  --config-scopes ukmo-restricted-scope
```

and it successfully generated md5 hashes of the `um7` package binaries.

### DIFF
--- a/scripts/release_provenance/get_data_from_deployment.py
+++ b/scripts/release_provenance/get_data_from_deployment.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import Any, List, Dict
 
 import spack.environment
+import spack.error
 import spack.cmd
 import spack.config
 import spack.paths
@@ -72,7 +73,11 @@ def add_custom_spack_config_scopes(config_scopes: List[str]) -> None:
 
     print(f"Attempting to load custom scopes: {config_scope_paths}")
 
-    spack.main.add_command_line_scopes(spack.config.CONFIG, config_scope_paths)
+    try:
+        spack.main.add_command_line_scopes(spack.config.CONFIG, config_scope_paths)
+    except spack.error.ConfigError:
+        print(f"Failed to find valid config scope in paths {config_scope_paths}.")
+        raise
 
 def activate_spack_environment(spack_env_path: str) -> spack.environment.Environment:
     spack_env = spack.environment.Environment(spack_env_path)
--- a/scripts/release_provenance/get_data_from_deployment.py
+++ b/scripts/release_provenance/get_data_from_deployment.py
@@ -20,7 +20,7 @@ import spack.main
 def main():
     args = parse_args(sys.argv[1:])
     packages: List[str] = args.packages.split(",")
-    config_scopes: List[str] = args.config_scopes.split(",")
+    config_scopes: List[str] = args.config_scopes.split(",") if args.config_scopes else []
     output_path = Path(args.output)
 
     # Custom scopes added via spack --config-scope for install need to be added back here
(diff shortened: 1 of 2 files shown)


scripts/release_provenance/get_data_from_deployment.py
```diff
@@ -50,6 +58,22 @@ def main():
...
 
+def add_custom_spack_config_scopes(config_scopes: List[str]) -> None:
+    """
+    Adds paths to custom spack config scopes to the command_line scope so we can find binaries for
+    certain environments that use custom installation directories.
+
+    :param config_scopes: Names of custom scopes from spack-configs custom/cd directory.
+    :type config_scopes: List[str]
+    """
+    spack_config_custom_scopes_path: Path = Path(spack.paths.spack_root).parent / "spack-config" / "custom" / "cd"
```
aidanheerdegen, line 70: Just a comment of future me: this is referring to this location in the speck-config repo clone:

https://github.com/ACCESS-NRI/spack-config/tree/main/custom/cd/

scripts/release_provenance/get_data_from_deployment.py
```diff
@@ -50,6 +58,22 @@ def main():
...
+
+    config_scope_paths: List[str] = [str(spack_config_custom_scopes_path / s) for s in config_scopes]
+
+    print(f"Found custom scopes: {config_scope_paths}")
```
aidanheerdegen: Have they been "found"? There is no verification that they exist.

CodeGat: True...updated in https://github.com/ACCESS-NRI/build-cd/pull/357/commits/1c9ecddd9a3ab4b1b208c70afd6ab00a696c5ef9

aidanheerdegen: Well .. is it worth doing just the merest of checks to ensure the environment directory exists? Otherwise it will silently fail.

CodeGat: :laughing: okay, got me there. 

Not silently - spack does it's own handling of that case, which I've added to a try/except in  https://github.com/ACCESS-NRI/build-cd/pull/357/commits/2dce4f8c265600b08b053b54e4bf9480660f0afe

aidanheerdegen (approved):
Sweet-as bro

PR #358, opened 2026-02-16
Correct spack v1.1 builtin-spack-packages to spack-config version

CodeGat:
## Background

For some reason, `spack v1.1` Pre/Release is using an older version of `builtin-spack-packages`. This PR updates it to be in line with `spack-config`s values for it. 

Future updates to `build-cd` mean that this `builtin-spack-packages` key will be removed, as we will be informed by `spack-config` for the `builtin-spack-packages` repo.

### DIFF
--- a/config/settings.json
+++ b/config/settings.json
@@ -14,7 +14,7 @@
         "1.1": {
           "spack": "2e2169d5282d166f63e3ee4db8d4446c43cefa8a",
           "spack-config": "2026.02.001",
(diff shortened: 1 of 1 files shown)


harshula (approved):

harshula:
Hi @CodeGat , Looks like the incorrect hash was introduced in PR https://github.com/ACCESS-NRI/build-cd/pull/352 and it had not reviewers. Regardless, my understanding is that `builtin-spack-packages` is going to be removed soon.

PR #359, opened 2026-02-16
`!update-configs` Comment Command Fixes

CodeGat:
Closes #354
Closes #341

## Background

The `!update-configs` comment command has been pretty useful so far, but there are a couple of issues aluded to in the above issues. Namely:

* The updated `config.yaml` has it's keys sorted, and removes comments, leading to a large diff
* New branches are created for every invocation of `!update-configs`, of the form `auto/prX-Y/CONFIG`, for each prerelease build

The fixes in this PR address both of these, respectively:

* We now use `ruamel` over `PyYaml`, which preserves comments and doesn't sort keys by default
* We now append commits to the `auto/prX/CONFIG` branch if a PR already exists. Note the lack of a `-Y` in `prX-Y` - we now create branches based on MDR PRs, not individual builds

## The PR

- **Use ruamel instead of PyYaml to preserve order/comments**
- **Append commits to existing auto/prX/CONFIG PR if they are from the same MDR PR**

### Testing of the script

Testing of the script itself was done on the `config.yaml` of https://github.com/ACCESS-NRI/access-om3-configs/blob/release-MC_25km_jra_iaf/config.yaml, giving the following diff, mostly with whitespace changes:

<details>
<summary>Large diff in this section</summary>

```diff
@@ -9,19 +9,19 @@
 # Using payu sync is recommend to copy data from ephemeral scratch space to 
 # longer term storage
 sync:
-    enable: False # set path below and change to true
-    path: null # Set to location on /g/data (e.g. /g/data/PROJECT/USER/EXPERIMENT)
-    restarts: True
-    exclude_uncollated: False
+  enable: false   # set path below and change to true
+  path: null   # Set to location on /g/data (e.g. /g/data/PROJECT/USER/EXPERIMENT)
+  restarts: true
+  exclude_uncollated: false
 
 #Model software version
 modules:
-    use:
-        - /g/data/vk83/modules
-    load:
-        - access-om3/2025.08.001
-        - model-tools/mppnccombine-fast
-
+  use:
+  - /g/data/vk83/modules
+  - /g/data/vk83/prerelease/modules
+  load:
+  - access-om3/prX-Y
+  - model-tools/mppnccombine-fast
 queue: normalsr
 ncpus: 2704
 jobfs: 10GB
@@ -32,34 +32,34 @@ jobname: 25km_jra_iaf
 model: access-om3
 exe: access-om3-MOM6-CICE6
 input:
[...]

PR #360, opened 2026-02-16
`!update-configs` Comment Command Fixes - `v7` Backport

### LINKED ISSUE #341: `!update-configs`: Updated Configs `metadata.yaml` Is Sorted, Removes Comments
Found by @dougiesquire in https://github.com/ACCESS-NRI/access-om3-configs/pull/1041/changes

## Background

Currently, `yaml.safe_dump()` sorts the yaml keys by default. In our case, this leads to a large diff. 

Furthermore, comments are also removed, which also contributes to this large diff. 

## Implementation Details

* For keeping keys in the original order, use `yaml.safe_dump(..., sort_keys=False)`
* For keeping comments, one may have to move from `PyYaml` to `ruamel`

### LINKED ISSUE #354: Rerunning `!update-configs` opens new PRs
The `!update-configs` command opens a new set of PRs every time it is run. It would be nice if re-running the same `!update configs` command instead updated the existing PRs, though as I'm writing this I realise this could be quite difficult with the profiles functionality...

CodeGat:
References #354
References #341
Backported from #359

## Background

The `!update-configs` comment command has been pretty useful so far, but there are a couple of issues aluded to in the above issues. Namely:

* The updated `config.yaml` has it's keys sorted, and removes comments, leading to a large diff
* New branches are created for every invocation of `!update-configs`, of the form `auto/prX-Y/CONFIG`, for each prerelease build

The fixes in this PR address both of these, respectively:

* We now use `ruamel` over `PyYaml`, which preserves comments and doesn't sort keys by default
* We now append commits to the `auto/prX/CONFIG` branch if a PR already exists. Note the lack of a `-Y` in `prX-Y` - we now create branches based on MDR PRs, not individual builds

## The PR

- **Use ruamel instead of PyYaml to preserve order/comments**
- **Append commits to existing auto/prX/CONFIG PR if they are from the same MDR PR**

## Testing

See https://github.com/ACCESS-NRI/build-cd/pull/359

### DIFF
--- a/.github/workflows/ci-command-configs.yml
+++ b/.github/workflows/ci-command-configs.yml
@@ -230,7 +230,7 @@ jobs:
           echo "Checking PR branch $pr_branch_name from ${{ env.CONFIG }}..."
           git -C caller-configs checkout ${{ env.CONFIG }}
 
-          if git -C caller-configs ls-remote --exit-code --heads origin $pr_branch_name; then
+          if git -C caller-configs ls-remote --exit-code --heads origin $pr_branch_name &> /dev/null; then
             git -C caller-configs checkout $pr_branch_name
           else
             git -C caller-configs checkout -b $pr_branch_name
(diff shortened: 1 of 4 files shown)


atteggiani (dismissed):
Looks good! 

There are minor implementation comments in [the related PR](https://github.com/ACCESS-NRI/build-cd/pull/359#pullrequestreview-3810870241) that are also valid for this PR.

Feel free to ignore them, or apply them here as well.

atteggiani (approved):

Issue #363, opened 2026-03-10
Can't generate metadata for `um` due to no `git` ref


CodeGat:
See https://github.com/ACCESS-NRI/ACCESS-rAM3/actions/runs/22880386731/job/66381741254?pr=24#step:18:49

## Background

Currently we use the `git` attribute to store a URL to a given package for the release database. However, some packages aren't stored on `git`, for example, we have some stored on `svn` and some on local repositories - like the `um`. 

## Implementation Plan

* ~~Check if the `um` package will move to `git` anytime soon.~~ It isn't
* ~~If not, determine what to do for packages without a `git` attribute. I think it was noted that all packages should be stored on `git` if possible - sound familiar, @aidanheerdegen ..?~~ Wait, I just remembered that `um` is special and has refs from `um_sources_ref` and `jules_sources_ref`. In that case...
* Get the ref specially for `um` from the `um_sources_ref` variant.

CodeGat:
But for rAM3 we don't have a `um_ref` as it is still using the Gadi svn mirror, which is not queryable to us. 
Maybe we should remove the `um` from `rAM3` until it has provenance via GitHub...?

PR #364, opened 2026-03-10
Get repo information specially for `um`

CodeGat:
See #363

## Background

We can't get the url or ref information for the `um` spack package because it doesn't have a `git` attribute in the `package.py`. It has special variants and data within the file for that purpose. 

## The PR

* Update the `get_packages_metadata()` function to get the `um` data specially - as in, it gets the URL from the `_resource_cfg` attribute at https://github.com/ACCESS-NRI/access-spack-packages/blob/38fec2deecd505a244ea47416eb1def01e4b0aa2/spack_repo/access/nri/packages/um/package.py#L183; and it's version from the value of the `um_ref` variant at https://github.com/ACCESS-NRI/access-spack-packages/blob/38fec2deecd505a244ea47416eb1def01e4b0aa2/spack_repo/access/nri/packages/um/package.py#L78

## Testing

Tested via https://github.com/ACCESS-NRI/ACCESS-AM3/pull/30

### DIFF
--- a/scripts/release_provenance/get_data_from_deployment.py
+++ b/scripts/release_provenance/get_data_from_deployment.py
@@ -102,6 +102,8 @@ def generate_packages_metadata(package_names: List[str], root_spec: spack.spec.S
 
         package_hash: str  = package.format('{hash}')
         package_location: str = package.format('{prefix}')
+        package_repo_url: str
+        package_repo_version: str
         package_repo_url, package_repo_version = _get_package_repo_info(package)
 
         md5s_of_binaries = generate_md5s_for_package_binaries(package)
--- a/scripts/release_provenance/get_data_from_deployment.py
+++ b/scripts/release_provenance/get_data_from_deployment.py
@@ -129,7 +129,7 @@ def _get_package_repo_info(package: spack.spec.Spec) -> Tuple[str, str]:
     if package_name == "um":
         return (
             spack.repo.PATH.get_pkg_class(package_name)._resource_cfg.um_ref.git_url,
-            package.variants.get("um_sources_ref").value
+            package.variants.get("um_ref").value
         )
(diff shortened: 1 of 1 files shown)


scripts/release_provenance/get_data_from_deployment.py
```diff
@@ -101,15 +101,14 @@ def generate_packages_metadata(package_names: List[str], root_spec: spack.spec.S
             raise
 
         package_hash: str  = package.format('{hash}')
-        package_version: str = package.format('{version}')
         package_location: str = package.format('{prefix}')
-        package_repo_url: str =  spack.repo.PATH.get_pkg_class(package_name).git
+        package_repo_url, package_repo_version = _get_package_repo_info(package)
```
jo-basevi, line 107: Did you still want to have `str` type hints for package_repo_url, package_repo_version?

jo-basevi (dismissed):
Looks good to me

scripts/release_provenance/get_data_from_deployment.py
CodeGat: It would be nice, but I don't know if you can do type hints for tuple unpacking? Unless we set the two variables to None first (with type hints)

CodeGat: After reading https://peps.python.org/pep-0526/#global-and-local-variable-annotations, I think I've got type hints for tuple unpacking done in https://github.com/ACCESS-NRI/build-cd/pull/364/commits/f26e5de27cd43d9d1d65c698b93131d587f03c8a

jo-basevi: Nice

jo-basevi (approved):
LGTM

PR #367, opened 2026-03-30
Added spack module

atteggiani:
Added a first instance of a spack module for deployment on Gadi. 

This PR is mainly to keep track of the modulefile for `spack` and allow versioned modification. It will be kept as a draft until we finalise the location within the repo where to place it and are happy to consider it merged into the infrastructure.


tools/modules/spack/.common
```diff
@@ -0,0 +1,150 @@
+#%Module1.0
+
+# Get module name and version
+lassign [split [module-info name] {/}] module_name module_version
+
+conflict $module_name
+
+# Set spack root
+set base_dir "/g/data/vk83/testing/apps/spack/$module_version"
```
CodeGat: I know this was just for the testing instance, but in the future, will there be one of these modulefiles in this repo for each version of `spack` we support? Or is there a way to make it more generic?

tools/modules/spack/.common
```diff
@@ -0,0 +1,150 @@
...
+        # If the old_env_var_name env variable doesn't exist, print out a warning message
+        puts stderr [string cat \
+            "WARNING: The $old_env_var_name env variable which stores the environment " \
+            "information prior to loading the $module_name module is not set. Environment restoration failed."
+        ]
+    }
+}
+
+# Set 'spacktivate' alias
+set-alias spacktivate {spack env activate}
```
CodeGat: Why do we need to set this alias specifically? 🤔

tools/modules/spack/.common
```diff
@@ -0,0 +1,150 @@
...
+            name="${entry%%%%=*}"
+            value="${entry#*=}"
+            before_variables["$name"]="$value"
+        fi
+    done < <(env -0)
+
+    # Set up SPACK_PYTHON
+    export SPACK_PYTHON="%s"
+    # Set up SPACK_ROOT
+    export SPACK_ROOT="%s"
```
CodeGat: I'm a bit confused with this part - is SPACK_PYTHON and SPACK_ROOT both being set to the same particular value..?

CodeGat (commented):
Just some comments

tools/modules/spack/.common
atteggiani: There are multiple approaches we can adopt, but what in particular did you want to make more generic? 

I don't know all the details of how the deployment works within this repo, but in general I would suggest we keep only one modulefile at each time within the repo.

Then, at the time of deployment, we can deploy the specific modulefile named according to its version (e.g., `1.0` for spack version `1.0`).
I prefer the solution above (having a completely separate modulefile for each spack version - even if they end up being identical for subsequent versions) rather than using a "symlinked common modulefile" approach (similar to the conda/analysis3 environments) because I don't see the benefits of it.

But I am happy to be convinced otherwise, and I think any solution can be discussed.

tools/modules/spack/.common
atteggiani: That alias is also [set by spack's `setup-env.sh`](https://github.com/spack/spack/blob/dce16ac37577f705033f360581a39987e109ced8/share/spack/setup-env.sh#L252), so I thought it would be good to replicate the behaviour of sourcing the file as much as possible.

tools/modules/spack/.common
atteggiani: No, they are set to their respective `$spack_python` and `$spack_root` values defined above.

If you notice the entire bash script is within a `format` TCL function (I give some details in the comments at line 32).

In general this is needed because we need to pass some TCL variables to the bash script. 
However, this can be highly simplified by passing them as arguments to the bash script, rather than using the `format` function. In fact, I think it should be done this way. I'll update the script.

Also, as suggested at line 36, the bash script could be separated from the modulefile (and deployed together with it). This would make things a bit cleaner and more easily maintainable.
If you'd like, I can also address this.

tools/modules/spack/.modulerc
atteggiani, line 1: A comment on this file:

This is an example. 
We can decide to set the default version like this, or have another way for setting it. 

If we choose this method, this file in the repo should probably be dynamic, with the version set at deployment time.

Other solutions (which would not require a dynamic file) could be:
- set the default version to the latest one deployed (if we're sure we always want the latest-deployed version to be the default one)
- default version automatically set based on version name (in this case we need a stable version naming scheme, and I think we do)

CodeGat:
Just noting that we will need to incorporate https://github.com/ACCESS-NRI/build-cd/pull/375 into the final design of the modulefile

atteggiani:
- Change python version to the one currently (3.6)
- Set base_dir dynamically based on the parent directory of the modulefile
- Separate bash script

CodeGat:
After a meeting with @harshula and @atteggiani :

* Rather than separating modules via Pre/Release based on the `module load` (for example, `module load spack/1.1/prerelease`/`module load spack/1.1/release`), we can encode that in the `module use`, like so:
  ```bash
  module use /g/data/vk83/modules
  module load spack/1.1
  # Loads the release instance of spack at /g/data/vk83/apps/spack/1.1 by virtue of the module location
  module use /g/data/vk83/prerelease/modules
  module load spack/1.1
  # Loads the prerelease instance of spack at /g/data/vk83/prerelease/apps/spack/1.1 by virtue of the module location
  ```
  * Furthermore, we should direct regular users to the `module use /g/data/vk83/modules` version. The prerelease version might be useful for getting alpha testers of new versions of `spack-config` or something. 

* In `.common`, if we can get the path to the modulefile itself, we can do some path magics to construct the `base_dir` - it'd probably look something like `$path_to_module_location/../../apps/spack/$module_version`

* If we could export `SPACK_USER_CACHE_PATH` in the module, that would be great. However, there are two cases to take into account:
  * The service user (has `SPACK_ADMIN_CONFIG` exported), which needs `SPACK_USER_CACHE_PATH=$base_dir/spack-admin-cache`
  * The regular user (does NOT have `SPACK_ADMIN_CONFIG` exported), which needs `SPACK_USER_CACHE_PATH=/g/data/$PROJECT/$USER/spack/$module_version/spack-user-cache`
  
Did I miss anything? :laughing:

atteggiani:
### Updates
The discussed updates have been implemented:
- [x] Spack python changed to be using the default python3 executable on Gadi (`/bin/python3`)
  The modulefile has been renamed to `modulefile`
  https://github.com/ACCESS-NRI/build-cd/pull/367/commits/6b893ee64467225b105dd7c57d6b39a4605e3aeb

- [x] The base directory and `spack_root` are now set dynamically based on the modulefile parent directory
  https://github.com/ACCESS-NRI/build-cd/pull/367/commits/9d9aabeb430007b1cee5d23310d603e72ea01f72

- [x] The bash script has been moved into separate file (`bash_file.sh`) for better clarity and maintainability.
  https://github.com/ACCESS-NRI/build-cd/pull/367/commits/a2be738fb00ea2172fba4b19101a1fdbcfaadb46

- [x] The spack user cache environment variable (`SPACK_USER_CACHE_PATH`) is exported based on the existence of `SPACK_ADMIN_CONFIG` env variable. 
  https://github.com/ACCESS-NRI/build-cd/pull/367/commits/d9c0b08f82bde48ab98373ebc7484dd1badbf404
  

### Info for the deployment
For simplicity, the modulefile and bash script in this repo have been named `modulefile` and `bash_script.sh`, respectively. 
At deployment time (within the build-cd deployment CD) the following should happen to match the current implementation:
- The deployed modulefile name should match the spack version deployed (e.g., if we're deploying spack version `1.1`, the modulefile should be named `1.1`). 
  An example modulefile path could be `/g/data/vk83/modules/spack/1.1`.
- The bash file should be deployed in the same directory as the modulefile, named as `.<version>.sh` (e.g., if we're deploying spack version `1.1`, the bash script should be named `.1.1.sh`). 
  An example bash script path could be `/g/data/vk83/modules/spack/.1.1.sh`.
  

### Testing
I copied the files (properly named) in `/g/data/vk83/prerelease/modules/spack` for testing, so this updated spack module can be used with:
```
module use /g/data/vk83/prerelease/modules/spack
module load spack/1.1
```

tools/modules/spack/modulefile
```diff
@@ -0,0 +1,89 @@
...
+            "information prior to loading the $module_name module is not set. Environment restoration failed."
+        ]
+    }
+}
+
+# Set 'spacktivate' alias
+set-alias spacktivate {spack env activate}
+
+# Set spack user cache path
+set spack_admin_config_env_var_name SPACK_ADMIN_CONFIG
```
CodeGat: Sorry @atteggiani, I got the environment variable name wrong - it's actually `ACCESS_SPACK_ADMIN`

tools/modules/spack/modulefile
```diff
@@ -0,0 +1,89 @@
...
+    #   When the module is unloaded, unset them.
+    # - When the module is loaded, set the variables and functions that changed to their new values.
+    #   When the module is unloaded, re-set them to their old values.
+
+    # To handle tcl and bash quoting correctly, we first set up the bash script in a TCL variable
+    # (bash_script) and then run it using `exec /bin/bash -c $bash_script`
+    # For this reason, we double the percent sign (%%) when we want a literal percent, because the script
+    # is within a TCL `format` function that would otherwise interpret the percent as a formatting sign.
+    # NOTE: this setup could be simplified by writing the bash script into a file 
+    # (shipped together with the spack module), and exec that file using somthing like `exec /bin/bash <path/to/file>`
```
CodeGat: Is this relevant anymore?

CodeGat (changes requested):

tools/modules/spack/modulefile
atteggiani: Updated in https://github.com/ACCESS-NRI/build-cd/pull/367/commits/bb1285866d3a60b74fd0a5ba1fa4d6cf319ebd8f

tools/modules/spack/modulefile
atteggiani: Nope :D deleted in https://github.com/ACCESS-NRI/build-cd/pull/367/commits/bb1285866d3a60b74fd0a5ba1fa4d6cf319ebd8f

CodeGat (dismissed):
Looks good! I might put an update to the `create-deployment-spack.yml` workflow that also deploys this - thanks!!

.github/workflows/create-deployment-spack.yml
```diff
@@ -74,3 +81,20 @@ jobs:
...
+            -i ${{ steps.ssh.outputs.private-key-path }} \
+            tools/modules/spack/bash_script.sh \
+            ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }}:${{ inputs.module-deployment-location }}/.${{ steps.strip.outputs.version }}.sh
+
+          rsync --archive \
+            -i ${{ steps.ssh.outputs.private-key-path }} \
+            tools/modules/spack/modulefile \
+            ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }}:${{ inputs.module-deployment-location }}/${{ steps.strip.outputs.version }}
+
+          echo "::notice::Modulefiles for ${{ steps.strip.outputs.version }} deployed. This does not affect the default module."
```
atteggiani, line 105: What does "This does not affect the default module." mean?

atteggiani:
@CodeGat The workflow changes for the deployment look good!
Just had a minor curiosity.

.github/workflows/create-deployment-spack.yml
CodeGat: As in, this workflow doesn't update the default `.modulerc` that says what the default module is

CodeGat (dismissed):

atteggiani:
### Updates n.2
The discussed updates have been implemented:
- [x] Spack python changed to be using `/bin/python3`

- [x] Added information on where the user spack cache is stored when the module gets loaded

- [x] Automatic detection of aliases (similarly to what is done for env vars and functions)

- [x] Restructure of the bash_script to be clearer and added logic to be able to reinstate vars, functions or aliases that get removed after sourcing the `setup-env.sh` (might never happen, but like this it's complete and we might not need to worry about updates of the `setup-env.sh` breaking the logic).


The updated files have been copied to `/g/data/vk83/prerelease/modules/spack`

harshula:
We should set `SPACK_PYTHON=/usr/bin/python3.12`, there is a significant performance improvement for concretization. See: https://github.com/ACCESS-NRI/build-cd/issues/394

CodeGat:
@atteggiani would you mind reviewing these last couple commits I made?

Essentially we will put the admin modulefiles under this structure:

```
.
└── vk83/
    ├── apps/
    │   └── spack/
    │       └── 1.1/
    │           ├── spack
    │           ├── spack-config
    │           └── ...
    ├── admin/
    │   └── modules/
    │       └── spack/
    │           └── 1.1
    └── prerelease/
        ├── apps/
        │   └── spack/
        │       └── 1.1/
        │           ├── spack
        │           ├── spack-config
        │           └── ...
        └── admin/
            └── modules/
                └── spack/
                    └── 1.1
```

So I needed to update the `modulefile-admin` to go an extra directory back in the case of the admin ones :)

CodeGat (dismissed):

.github/workflows/create-deployment-spack.yml
```diff
@@ -74,3 +86,20 @@ jobs:
...
+      - name: Deploy modulefiles
+        run: |
+          rsync --archive \
+            -i ${{ steps.ssh.outputs.private-key-path }} \
+            tools/modules/spack/bash_script.sh \
+            ${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }}:${{ inputs.module-deployment-location }}/.${{ steps.strip.outputs.version }}.sh
+
+          rsync --archive \
+            -i ${{ steps.ssh.outputs.private-key-path }} \
+            tools/modules/spack/${{ inputs.module-deployment-is-admin && 'modulefile-admin' || 'modulefile' }} \
```
atteggiani: This could be simplified to:
```suggestion
            tools/modules/spack/modulefile${{ inputs.module-deployment-is-admin && '-admin' }} \
```

atteggiani (commented):
Just a minor suggestion. Otherwise it looks good.
Thanks @CodeGat

CodeGat (approved):

PR #368, opened 2026-04-09
Updates for new `spack-config`

### LINKED ISSUE #349: Can `builtin-spack-packages`/`access-spack-packages` be informed solely by `spack-config`?
## Background

There are cases where:

* We update the `builtin-spack-packages` ref in `build-cd`s `config/settings.json` for Release and Prerelease, and
* We update the `access-spack-packages` ref in a MDRs `config/versions.json` for Release,
 
which is then updated in the spack instance. However, this leads to the `repos.yaml` coming out of sync with the `spack-config` (becoming modified by the `spack-config` repo), which is then blown away when `spack-config` is updated. 

### Change config scope for `builtin-spack-packages` updates

Have updates to `builtin-spack-packages` go to a higher configuration scope, so they are not considered modified in the `spack-config` repository, but still take effect.

However, this will mean that `spack-config`s `repos.yaml` versions will be able to be overridden.

### Have only `spack-config` inform `builtin-spack-packages`

Alternatively, don't have the `builtin-spack-packages` key in `config/settings.json` at all - just as it was beforehand. In this case, `spack-config` (rather than `spack`) will denote the version of `builtin-spack-packages`. More testing will need to be done regarding how to update the versions of the packages when `spack-config` is updated.

However, this will remain an issue for Releases for `access-spack-packages` - `spack-config` will be modified when an MDRs `access-spack-packages` is updated in `config/versions.json`.

CodeGat:
References ACCESS-NRI/spack-config#41
Closes https://github.com/ACCESS-NRI/build-cd/issues/349

## Background

There is an update to `spack-config` (see https://github.com/ACCESS-NRI/spack-config/pull/110), which makes use of the `includes.yaml`, and a move away from the `spack-enable.bash`. This PR updates the workflow to use `setup-env.sh`, and adds a new script to set up service users. 

Furthermore, we revert to essentially pre-`v8` behavior in which instead of `builtin-spack-packages` being a lever to pull in `config/settings.json`, it is informed solely by `spack-config`.

## The PR

- **Swap from spack-enable.bash to setup-env.sh**
- **Add tools/service-user for setup of service users, update README**
- **Remove updating of `builtin` via `settings.json`, now informed by `spack-config`**

## Testing

Tested in Prerelease via https://github.com/ACCESS-NRI/ACCESS-TEST/pull/65

### DIFF
--- a/.github/workflows/deploy-2-start.yml
+++ b/.github/workflows/deploy-2-start.yml
@@ -197,6 +197,10 @@ jobs:
           # Export vars.DEPLOYMENT_TARGET
           export DEPLOYMENT_TARGET="${{ vars.DEPLOYMENT_TARGET }}"
           echo "DEPLOYMENT_TARGET exported as $DEPLOYMENT_TARGET"
+
+          # This environment variable is used by our spack-config to delineate between a normal user
+          # and an admin or service user that has custom config.install_trees (among other things).
+          # See https://github.com/ACCESS-NRI/spack-config/blob/main/v1.1/include/defaults.yaml
           export ACCESS_SPACK_ADMIN=1
 
           # Enable spack
@@ -244,6 +248,10 @@ jobs:
           # Export vars.DEPLOYMENT_TARGET
           export DEPLOYMENT_TARGET="${{ vars.DEPLOYMENT_TARGET }}"
           echo "DEPLOYMENT_TARGET exported as $DEPLOYMENT_TARGET"
+
+          # This environment variable is used by our spack-config to delineate between a normal user
+          # and an admin or service user that has custom config.install_trees (among other things).
+          # See https://github.com/ACCESS-NRI/spack-config/blob/main/v1.1/include/defaults.yaml
           export ACCESS_SPACK_ADMIN=1
 
           . ${{ steps.path.outputs.spack }}/share/spack/setup-env.sh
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy-2-start.yml
```diff
@@ -156,6 +156,15 @@ jobs:
...
 
+      - name: Validate service user permissions
+        run: |
+          ssh ${{ secrets.USER }}@${{ secrets.HOST }} -i ${{ steps.ssh.outputs.private-key-path }} /bin/bash <<'EOT'
+            if [ -z "$ACCESS_SPACK_ADMIN" ]; then
+              echo "::error::Service user is not setup to deploy Pre/Releases. Please ensure the service user has the correct permissions and try again."
+              exit 1
+            fi
+          EOT
+
```
aidanheerdegen: I understand why, but I don't think we should do this check here seeing as we explicitly set this later.

If you want to make sure this is set I'd suggest creating a cron job to do the check for your own sanity.

CodeGat: Done in https://github.com/ACCESS-NRI/build-cd/pull/368/commits/853d1ef3d29f8e6328a4bed8d836a2e061489b5c

CodeGat:
@CodeGat don't forget to update the hash of spack in `config/settings.json` post-merge of https://github.com/ACCESS-NRI/spack-config/pull/110 ! 

EDIT: Done in https://github.com/ACCESS-NRI/build-cd/pull/368/commits/f0e4d61cb182d7c2b20c51e998b76de9ffd7ba57

harshula (approved):
Reviewed via Zoom VC

Issue #369, opened 2026-04-13
`restricted/ukmo` modules generated without scope


CodeGat:
## Background

In the `build-cd v8` changes, we attempted to scope the regeneration of modules to the `restricted/ukmo` scope (if required). This didn't work. In #XXX, we returned it to it's original behavior. 

Originally, we generated modules for packages that are in the `restricted/ukmo` scope, without a scope. This means that people can attempt to use/load models that they aren't allowed to use. It will fail for users that don't have the permissions, but this isn't a satisfying solution. 

## Implementation Plan

* In the `restricted/ukmo` scope, create a custom `modules.yaml` that sets the root to `$spack/../restricted/ukmo/release/modules`, so we aren't writing to the unscoped area (eg. `$spack/../release/modules`)

Pinging @aidanheerdegen

aidanheerdegen:
Note that this was attempted without a corresponding custom `modules.yaml` which broke the pre-release deployment `module load`, and this was reverted in this PR https://github.com/ACCESS-NRI/build-cd/pull/370

PR #370, opened 2026-04-13
Regenerate Modules Without Custom Scopes

CodeGat:
References #369.

## Background

For models with restricted scopes, we both `spack install` and `spack module tcl refresh` using thoes scopes. This doesn't work for the module refresh bit.

This PR reverts changing the scope of `spack module tcl refresh`, allowing modules to be successfully loaded again.

## The PR

* Remove loading of custom scopes during `spack module tcl refresh`

## Testing

Tested locally by regenerating the modulefiles of a Prerelease using `spack -e access-am3-pr34-1 module tcl refresh -y`. It worked!


aidanheerdegen (approved):
LGTM

PR #373, opened 2026-04-14
deploy-2-start: Add --upstream-modules to module refresh code

CodeGat:
## Background

Spack added a `--upstream-modules` flag to `spack module tcl refresh` that refreshes modules defined in an upstream - this is the case in our `ukmo-restricted` scope, which makes `$spack/../release` an upstream of our `$spack/../restricted/ukmo/release` dir. This means that we can now generate modulefiles for externals defined in an upstream (like `openmpi`).
 
## The PR

* Add `--upstream-modules` flag.


aidanheerdegen (approved):
LGTM. Nice find!

PR #375, opened 2026-04-17
Make service user cache a sibling of spack

CodeGat:
References https://github.com/ACCESS-NRI/spack-config/issues/41

## Background

The user cache path defaults to `~/.spack`. This is probably fine for regular users, but for service users, who manage multiple spack instances, we don't want to mix configuration of any kind between Pre/Release instances, or different versions of `spack`. 

Therefore, we should explicitly set `SPACK_USER_CACHE_PATH` (similar to what the old `spack-enable.bash` did) in the workflow, based on the spack instance path. In the future, if there is an easy way to set it in `spack-config`, we will move to that. 

## The PR

* Export `SPACK_USER_CACHE_PATH` as a sibling of the `spack` instance for steps where we use `spack`.


CodeGat:
Closing as this will be incorporated into https://github.com/ACCESS-NRI/build-cd/pull/367

PR #384, opened 2026-05-18
TEMP: Add step that adds timestamp and public runner ip to artifact on failure

CodeGat:
## Background

We're currently investigating an uptick in failures to connect via SSH. To have NCI assist us, we need the public IP of the runners that are attempting to SSH/rsync. This will be temporary and reverted once the ticket is closed.

## The PR

* Adds a step that uploads an artifact containing a timestamp and runner IP on failure of any step that uses `rsync`/`ssh`.


CodeGat:
Closing as this is no longer required

PR #385, opened 2026-05-19
Update Actions that use Node 20

CodeGat:
See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

## Background

Node 20 Actions will be forced to run on Node 24 in June. This PR updates all actions that use Node 20 to versions that use Node 24. Since they are `actions/*` actions, the major branch version is updated. 

## The PR

- **Update actions/checkout to v6**
- **Update actions/setup-python to v6**
- **Update actions/*-artifact to v7**

### DIFF
--- a/.github/actions/get-spack-manifest/README.md
+++ b/.github/actions/get-spack-manifest/README.md
@@ -26,7 +26,7 @@ jobs:
   manifest:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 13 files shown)


jo-basevi (approved):
LGTM

Any reason to not use `v8` for [actions/download-artifact](https://github.com/actions/download-artifact)?

CodeGat:
Mostly just so it had major version parity with `upload-artifact` :grin:

PR #390, opened 2026-05-20
Checkout actual `spack-config` ref supplied, update references to new `upstream-spack-packages` repository

CodeGat:
See failing workflow https://github.com/ACCESS-NRI/ACCESS-OM2/actions/runs/26136102887/job/76871574166?pr=148

## Background

The above workflow actually demonstrates two bugs:

### Bug 1: Using a non-existent step output

In `deploy-1-setup.yml`s `Checkout spack-config to get builtin spack-packages ref` step, we mistakenly template `actions/checkout`s `ref` with `steps.config-settings-refs.outputs.spack-config` (which doesn't exist - should be `steps.config-settings-refs.outputs.spack-config-ref`). This means that it has been using the current `main` of `spack-config` for validation of configuration, rather than the version that the spack instance is currently using. 

Due to a recent change (https://github.com/ACCESS-NRI/spack-config/pull/113) to `spack-config`, it led to us attempting to validate a `ACCESS-NRI/upstream-spack-packages` branch against the `spack/spack-packages` repository. Which leads to the second bug...

### Bug 2: Not updating references to `upstream-spack-packages`

We have moved to `ACCESS-NRI/upstream-spack-packages`. But the CI is still validating against the original `spack/spack-packages`. Even if the first bug hadn't happened, this would show up when we updated the spack instances `spack-config` to the current `main`. 

## The PR

* Checkout the actual `spack-config` ref, rather than `main`, by using the proper template variable
* Update all references of `spack/spack-packages` to `ACCESS-NRI/upstream-spack-packages`.

### DIFF
--- a/.github/workflows/deploy-1-setup.yml
+++ b/.github/workflows/deploy-1-setup.yml
@@ -109,11 +109,11 @@ on:
     #   builtin-spack-packages-version:
     #     value: ${{ jobs.check-config.outputs.builtin-spack-packages-version }}
     #     description: |
-    #       Git ref of 'spack/spack-packages' that is used to reference builtin packages in spack.
+    #       Git ref of 'ACCESS-NRI/upstream-spack-packages' that is used to reference builtin packages in spack.
     #   builtin-spack-packages-git-hash:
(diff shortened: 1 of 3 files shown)


aidanheerdegen (approved):
LGTM

PR #397, opened 2026-06-10
PR cleanup: keep build dependencies when garbage collecting

CodeGat:
## Background

During manual garbage collection of the release instance, I noticed that build-time dependencies would be cleaned up with `spack gc --except-any-environment` - including things like `cmake`. Maybe this is contributing to the lack of reuse of low-level dependencies in Prerelease?

When I added `--keep-build-dependencies`, it kept those installs of `cmake`, etc. Maybe it's worth keeping those around for a while to see if it aids reuse?

## The PR

* When cleaning up PRs in the Prerelease instance, `--keep-build-dependencies` so hopefully installs of `cmake` and other low-level deps are reused in Prerelease.

### DIFF
--- a/.github/workflows/undeploy-1-start.yml
+++ b/.github/workflows/undeploy-1-start.yml
@@ -82,12 +82,12 @@ jobs:
           done
 
           # Base scope cleanup (most environments are in this scope)
-          spack gc --except-any-environment --yes-to-all
+          spack gc --except-any-environment --keep-build-dependencies --yes-to-all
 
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
> When I added `--keep-build-dependencies`, it kept those installs of `cmake`, etc. Maybe it's worth keeping those around for a while to see if it aids reuse?

What is contained in "etc"? Can we quantify what the inode/storage impact would be?

harshula (approved):
🤞

